PR #110, opened 2020-08-29
restore sticky?

beshaya:


beshaya:
Did it with javascript in another PR.

PR #118, opened 2020-08-29
Fix numerous calculator nits

### LINKED ISSUE #94: Put commas in numbers 1000 or bigger
image

It's hard to read "40000", I can't count the zeros.

there used to be a "toLocaleString()" that did this, but it seems to have gone away with the proper representation of sig figs, which is a worthy tradeoff but I do want the comma back!

oremanj:
- Display positive test rate as "no data available" if no data is available, not "20%"
- If the user selects a location with zero reported cases, show them zero risk instead of disabling parts of the calculator
- If we don't have enough info to calculate, show "Results: fill in calculator to see" instead of a string with a bunch of dashes in it
- Add commas to displayed microCOVID total - fixes #94
- Calculation methods now always return null if they have insufficient data, and we display that as "results: fill in to see"
  - Zero is not null
     - Really truly
  - Having a mask/volume/distance dropdown on "Select..." isn't considered insufficient -- we use the baseline value. (I'm not attached to this but I do think it's more usable.)
  - Insufficient = no interaction type selected, zero duration for one-off interaction, zero people, or (when using custom prevalence) zero population/cases.
  - If you decide to require something to be selected from every menu before a number is displayed, please make sure you fix the current bug where it would show a really high value (tens of thousands of uCoV) instead of dashes. It's gone away with this PR.
- Move "(most of the time)" text from distance choices to distance label
- Prevent nonnegative numeric fields from going below zero


catherio:
yes, not allowing manually entering 0 is WAI

PR #147, opened 2020-08-30
Implement sticky results bar in javascript.

beshaya:
Avoids needing to change overflow.

### DIFF
--- a/src/pages/Calculator.tsx
+++ b/src/pages/Calculator.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react'
+import React, { useEffect, useMemo, useState } from 'react'
 import { Col, Row } from 'react-bootstrap'
 
 import {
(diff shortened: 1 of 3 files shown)


src/pages/Calculator.tsx
```diff
@@ -26,6 +26,14 @@ const localStorage = window.localStorage
 const FORM_STATE_KEY = 'formData'
 
 export const Calculator = (): React.ReactElement => {
+  // Mount
+  useEffect(() => window.addEventListener('scroll', scrollListener, true), [])
+  // Unmount
+  useEffect(
```
dominic, line 32: One too many useEffects; just return the unmount handler from the first call

https://reactjs.org/docs/hooks-reference.html

src/pages/Calculator.tsx
```diff
@@ -207,16 +215,38 @@ export const Calculator = (): React.ReactElement => {
...
         <Col lg={{ span: 8, offset: 4 }}>
           <ExplanationCard points={points} />
         </Col>
       </Row>
     </div>
   )
 }
+
+// Sets #pointsRow to position: fixed if it would not be on the screen.
+function scrollListener(): void {
```
dominic, line 233: New function syntax preferred: `const scrollListener = ...`

dominic:
This is less helpful when scrolling down the calculator right now - maybe only enable stickiness if there's a point total available as well?

image

src/pages/Calculator.tsx
beshaya: Done!

Issue #137, opened 2020-08-29
Split into three mask types


catherio:
Split into three mask types:
* No mask
* Bandana, buff, or single-layer cotton
* Mask with filter insert, or N95 or better

aiui Bandana/buff is quite good for source control, not as good for wearer protection.

catherio:
Also add advanced masks from spreadsheet to calculator cc @beshaya 

(and explain all the above in the main text)

beshaya:
I'm taking this and will add outflow valves as well.

catherio:
For *this* issue, we just want

My mask protects me:
No mask - None
Bandana, buff, or single-layer cotton - None
Surgical mask, mask with filter insert, or N95 or better - 2x

Your mask protects me:
No mask - None
Bandana, buff, or single-layer cotton - 4x
Surgical mask, mask with filter insert, or N95 or better - 4x

catherio:
You could also do #159 
and #136

catherio:
Mask with outflow valve should not make any difference in this PR. It is as good as a basic mask.

beshaya:
I thought outflow valves are considered bad. Is the law against them misguided?

catherio:
Do you have any evidence about that?

All I have is this from the Fischer study
![image](https://user-images.githubusercontent.com/245037/91670731-07527700-ead5-11ea-8f1d-7164015604e5.png)

(the study everyone's using to freak out about bandanas, which, I also don't think is justified)

catherio:
Let's save valves for https://github.com/microcovid/microcovid/issues/159 which is labeled "epistemic deep dive"!

beshaya:
Fine deferring valves. 

That chart shows valved N95 significantly worse than fitted n95 or surgical for "their mask."

catherio:
more about the specific trifurcation suggested here:
![image](https://user-images.githubusercontent.com/245037/91670775-5bf5f200-ead5-11ea-93bf-053692318641.png)

blanchardjeremy:
Question: did this change also make it into the spreadsheet?

Issue #159, opened 2020-08-30
Explain mask types in much more detail


catherio:
Also add advanced masks from spreadsheet to calculator cc @beshaya 

(and explain all the above in the main text)

catherio:
Include valves, see discussion https://github.com/microcovid/microcovid/issues/137#issuecomment-683478300

catherio:
Review this resource as part of this investigation: http://built-envi.com/what-kind-of-mask-should-i-be-wearing-to-protect-against-covid-19/

catherio:
Review this resource too! http://jv.colostate.edu/masktesting/

catherio:
Obvious resource! https://www.preprints.org/manuscript/202004.0203/v3

catherio:
More resources: https://twitter.com/jljcolorado/status/1306323133599592448

catherio:
Another resource: https://twitter.com/jeremyphoward/status/1323278825275322370

catherio:
Connects to https://github.com/microcovid/microcovid/issues/405

catherio:
Connects to https://github.com/microcovid/microcovid/issues/340

catherio:
https://t.co/MGwaXSV9mI?amp=1

catherio:
https://t.co/MGwaXSV9mI?amp=1

PR #163, opened 2020-08-30
Update masks to trichotomy: none, cotton, filtered

beshaya:
Fixes: #137

Updates mask labels to be clear and user-friendly. Links from calculator to relevant research source section. Clearly lists cotton masks as "basic" masks, and reduces wearer protection from 2x to 1x for these.

I think this is sufficient for #159 (the wording is clear enough to answer most of the user questions we're seeing). I added a popover box to link to the relevant research section. Note that the research sources section cites Chu and states 2x wearer protection for basic masks while the calculator is using 1x.

### DIFF
--- a/src/components/calculator/ActivityRiskControls.tsx
+++ b/src/components/calculator/ActivityRiskControls.tsx
@@ -140,7 +140,7 @@ const maskPopover = (
   <Popover id="popover-basic">
     <Popover.Title as="h3">Masks</Popover.Title>
     <Popover.Content>
-      For more details on masks, see&nbsp;
+      For more details on masks, see{" "}
       <Link to="/paper/14-research-sources#masks" target="_blank">
         research sources
       </Link>
--- a/src/components/calculator/ActivityRiskControls.tsx
+++ b/src/components/calculator/ActivityRiskControls.tsx
@@ -1,4 +1,6 @@
 import React from 'react'
+import { Popover } from 'react-bootstrap'
+import { Link } from 'react-router-dom'
 
 import { SelectControl } from './SelectControl'
 import {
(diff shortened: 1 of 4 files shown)


src/components/calculator/ActivityRiskControls.tsx
```diff
@@ -131,3 +135,15 @@ export const ActivityRiskControls: React.FunctionComponent<{
     </React.Fragment>
   )
 }
+
+const maskPopover = (
+  <Popover id="popover-basic">
+    <Popover.Title as="h3">Masks</Popover.Title>
+    <Popover.Content>
+      For more details on masks, see&nbsp;
```
oremanj, line 143: `see{' '}` is more typically how we would force this space, since it doesn't need to be non-breaking (just needs to exist)

src/data/data.ts
```diff
@@ -111,22 +115,19 @@ export const Distance: { [key: string]: FormValue } = {
   sixFt: { label: '6+ feet apart', multiplier: 0.5 },
   tenFt: { label: '10+ feet apart', multiplier: 0.25 },
 }
+
+const noneLabel = 'No mask or poorly-worn mask'
+const basicLabel = 'Cotton mask, bandana, or buff'
+const filteredLabel = 'Surgical mask, mask with filter insert, or N95'
```
oremanj, line 121: I would remove "or N95" -- we strongly believe N95 is much better than 2x wearer protection, putting it in as 2x seems confusing to me

src/components/calculator/ActivityRiskControls.tsx
beshaya: Done

src/data/data.ts
beshaya: Done

PR #170, opened 2020-08-31
Fix crash with outdated local storage data

oremanj:
Hotfix part 1: don't crash if calculate() raises an exception


oremanj:
I used the JS console to set local storage with an invalid mask type, then verified upon refreshing that the calculator displays and that mask option is on "Select..."

PR #176, opened 2020-08-31
Fix part 2 for outdated local storage data: fix it up when loading it

oremanj:
This should hopefully prevent future issues like #170. It also adds a customization point where we can do smarter fixes than "change the corresponding field to nothing-selected-yet".


catherio:
(I'm not merging because I don't have time/attention right now to babysit the deployed site and step in if it fails)

catherio:
@oremanj  FYI this is ready to go in, since you seem to be responsive right now: I just don't have dev spoons to babysit :)

PR #202, opened 2020-09-02
Use percentage points (not %) as units of increase in risk

david-mears-dfe:
I am not totally sure that this is correct: I am not sure whether the number provided here is supposed to be an increase from e.g. 5% total risk to 15%, or from 5% to 5.5%. I assume that it is the former (percentage points), but haven't checked that 
the calculation (that I haven't touched) is a calculation of percentage points.

Whether or not I am correct, this paragraph should be made less ambiguous, as there is a big difference between 5.5% and 15% total risk, and people could interpret a 2% _relative_ increase as very small (missing that it is a 2% _absolute_ increase).

Screenshot 2020-09-02 at 14 48 07

Reviewers: if you like the change, please consider whether the sentence could be phrased more neatly or more clearly.

### DIFF
--- a/src/components/calculator/PointsDisplay.tsx
+++ b/src/components/calculator/PointsDisplay.tsx
@@ -29,7 +29,7 @@ function tooManyPoints(points: number): boolean {
 function pointsPerWeekToAnnual(points: number): string {
   return showPoints(points)
     ? fixedPointPrecisionPercent(1 - (1 - points * 1e-6) ** 52)
-    : '—%'
+    : '—'
 }
 
 function maybeGreater(points: number): string {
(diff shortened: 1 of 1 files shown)


david-mears-dfe:
Or how about:

> [...] you would have a 10% chance of getting COVID [...] (not including your risk from everything else you do!)

i.e. getting rid of the phrase 'an additional' (which I changed to 'more').

catherio:
Can you say more about why you think "more" is clearer than "an additional"?

beshaya:
Thanks for the PR, David! I agree that this block isn't super clear, I think we're going to just it in favor of a line about weekly budget: #206

david-mears:
> Can you say more about why you think "more" is clearer than "an additional"?

I meant that not saying 'more' OR 'an additional' is more accurate than saying either. My first draft used 'more'.

Issue #204, opened 2020-09-02
Compare total risk to risk budget


beshaya:
![image](https://user-images.githubusercontent.com/5877421/92034581-4c1f1d80-ed22-11ea-8497-40f2d0ecbecd.png)

Users are confused by small numbers. Help them compare to the weekly risk budget, but *don't* use "afford" terminology.

blanchardjeremy:
Question on this: maybe I'm misunderstanding the intent, but it seems like the calculation is off: 

![image](https://user-images.githubusercontent.com/32290/92261831-00d64d80-ee8f-11ea-9c7d-b9018a3b452a.png)


By my math 

`30/(10000/52) = 15.6%` 

Curious why the screenshot shows 20%

beshaya:
We only show 1 sigfig across the site.

PR #206, opened 2020-09-02
Compare total risk to risk budget.

beshaya:
image

image

Fixes: #204


linesarefuzzy:
As a user, just stumbling on this PR, I like the idea a lot, but I suggest adding it to the existing annual risk language rather than replacing it. I like seeing the annualized risk too.

catherio:
🤷‍♀️  I'm open to including both

beshaya:
Added the annualized risk back. I don't like the additional clutter, personally.

catherio:
agree, don't prefer the clutter. revert 7d6b55b and I'll LGTM?

beshaya:
@catherio Done

catherio (approved):
LGTM

PR #207, opened 2020-09-02
People do wear bandanas in grocery stores

catherio:
After our trichotomy update, I think grocery stores should look a bit less scary than this.


catherio:
oh, duh

oremanj (approved):

Issue #209, opened 2020-09-02
Person risks are inconsistent with stated scenario


beshaya:
The risks that are built off "living alone" (n housemates, closed pods, etc) rely on hard-coded multipliers. When we decreased the effectiveness of cloth masks, they got out of sync.

catherio:
Fix these in the same pass that includes #159

PR #211, opened 2020-09-02
Make risk profiles self-consistent, and use them as a regression test.

beshaya:
Also add a regression test for calculating regional person risk.

Fixes: #209

### DIFF
--- a/src/data/__tests__/calculate.test.ts
+++ b/src/data/__tests__/calculate.test.ts
@@ -77,7 +77,7 @@ describe('calculate', () => {
       distance: 'sixFt',
       duration: 60,
       theirMask: 'basic',
-      yourMask: 'basic',
+      yourMask: 'filtered',
       voice: 'silent',
     }
 
--- a/src/data/__tests__/calculate.test.ts
+++ b/src/data/__tests__/calculate.test.ts
@@ -1,20 +1,38 @@
-import { CalculatorData, calculate } from 'data/calculate'
+import {
+  CalculatorData,
+  calculate,
+  calculateLocationPersonAverage,
+  defaultValues,
+} from 'data/calculate'
+import { RiskProfile } from 'data/data'
 import { prepopulated } from 'data/prepopulated'
 
 describe('calculate', () => {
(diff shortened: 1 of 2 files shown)


oremanj (commented):
The cleanup to use named constants is great. It might be easier to read if you write the accesses like `Interaction.oneTime.multiplier` instead of with brackets and quotes.

The change to assume the people we’re modeling in the person risk scenarios use basic masks instead of filtered masks seems reasonable to me but I’d like catherio to OK it as well before we land. If you split that to a different PR, I’m happy to approve this one (which is then just the refactoring) immediately.

beshaya:
I changed the notation.

I consider this fixing a bug that was introduced in 6467c0d510d3398ee7d570f220f5038a9f3111bf - before that we said that all masks gave you a 2x protection factor. When we nerfed cotton, these values never got updated.

oremanj:
We didn’t specify the own-mask type for these models. Assuming it’s “basic” is just as arbitrary as assuming it’s “filtered”. Thus, my preference to leave the numbers as-is for right now.

catherio:
fwiw I don't feel confident of nerfing cotton; my intention with the trifurcation was to very quickly get out of a weird state where we were excessively downgrading buffs for source control (calling them "nothing", which is *super* not true afaict), but without also having to in that same edit stake a claim that buffs are great for protecting onesself (which seems dubious to me?). I don't know where the exact line *should* be and I'm deferring judgment until we do a masks overhaul. I just wanted to fix a Very Bad state quickly.

in other words, *this* fix feels like over-reifying and over-propagating a quick fix, rather than waiting for the full real good fix. I'm okay leaving this in a state where we assume your contact wore a surgical or fitted filtered mask, and randos in the store wear buffs and bandanas.

I don't want to push super-user-affecting changes all the time on minimal thought, would prefer to do so more slowly and with more consideration.

I also prefer to leave the numbers as-is for now.

beshaya:
Switched to filtered.

oremanj (approved):

PR #224, opened 2020-09-05
Fixed typo in word 'MODify'

richardgill:


oremanj:
This is intentional -- it's a reference to "MOD" = "Masked, Outdoors, and Distanced" as introduced in Section 3.

Issue #145, opened 2020-08-30
Handle regions with negative daily cases


Grubla:
Sweden and most of it's regions have a negative number of "Reported cases in past week". The total numbers for Sweden has declined in the JHU CSSE data, between 8/26 and 8/28 since the historic data has been corrected (the actual cases removed are over the last months, and were called false positives). 

This breaks the calculator and no result or risk level is given. 

![image](https://user-images.githubusercontent.com/1529750/91656502-70c48c80-eab9-11ea-9503-5a6f063f7dbc.png)

Grubla:
I couldn't find any negative numbers from ourworldindata, but I'd be willing to help/investigate if someone points me in the right direction :)

oremanj:
Thanks for the report! We're using JHU CSSE because it was the only source we could find for international data that was finer-grained than an entire country. Unfortunately they don't go back and update their previous "total new cases as of <date>" reports when historical corrections are made, which sometimes leads to issues like this. (Last week it was Massachusetts.)

If you have another source of prevalence data, you can still use the calculator by entering it in the boxes manually. And I bet the negativity will resolve itself when we pull updated data in a week or so. I'm not sure if we'll be able to add the ability to fall back to an alternate data source before then, but I'll update this issue if we do so.

oremanj:
This is resolved in the latest data update. We still don't have a principled solution, but hopefully we won't wind up needing one.

PR #228, opened 2020-09-07
Update prevalence data

### LINKED ISSUE #227: County data not updating
I tried a couple of different browsers in case the information was just cached. But no matter how many times I reload the page, I still see "Data last updated August 28, 2020." Our county has been spiking as many others' have and it would be great not to have to calculate the county info by hand. Is this supposed to update automatically? Or am I just doing it wrong?

oremanj:
Fixes #227, #145


update_prevalence.py
```diff
@@ -621,13 +621,29 @@ def main() -> None:
...
 
+        # Hack: On 2020-09-01, JHU switched from reporting all NYC cases
+        # under a virtual county called "New York City" to reporting them
+        # under the counties for individual boroughs. Until 2020-09-14, we'll
+        # support mixing the pre- and post-transition data by merging the
+        # five borough counties (New York, Kings, Queens, Richmond, Bronx)
+        # into "New York City".
+        nyc_county_names = ("New York", "Kings", "Queens", "Richmond", "Bronx")
+        ny_state = data.countries["US"].states["New York"]
+        nyc_indv = [ny_state.counties.pop(cname) for cname in nyc_county_names]
```
beshaya, line 640: Doesn't pop() remove the county from the dict? The above comment seems to imply that this week is meant as a transition in which we should expose the county data.

Also... why support New York City as a county for a limited amount of time? Seems if you're going to drop it in the future, why not now?

update_prevalence.py
```diff
@@ -621,13 +621,29 @@ def main() -> None:
...
+        nyc_combined = ny_state.counties["New York City"]
+        if not nyc_combined.population:
+            nyc_combined.population = sum(c.population for c in nyc_indv)
+        for county in nyc_indv:
```
beshaya, line 644: Use a list comprehension and sum() for consistency with above?

update_prevalence.py
oremanj: Before 2020-09-01, we only have aggregated data for NYC. After 2020-09-01, we only have split-out data per borough. Thus, the only thing we can report for a time interval that spans the transition is the aggregated data. Thus, pop() to remove the individual borough counties is correct. (Otherwise, you could select an individual borough in the dropdown but it wouldn't have a full week's worth of cases.)

update_prevalence.py
oremanj: Some of the cases are already filed under `nyc_combined`, so it's a different situation than for population. I guess we could write `nyc_combined.cumulative_cases += sum([c.cumulative_cases for c in nyc_indv], collections.Counter())` or `nyc_combined.cumulative_cases = sum([c.cumulative_cases for c in nyc_indv], nyc_combined.cumulative_cases)` but either of those seems harder to understand to me.

update_prevalence.py
beshaya: Ah, ok, so this code will be removed on the 14th? Could you put in an early return on that date so that this can't get rolled forward by accident?

update_prevalence.py
beshaya: Wow, that's confusing. Consider leaving a comment saying that some are populated.

beshaya (approved):
LGTM % small comment requests

PR #258, opened 2020-09-15
Use absolute path for opengraph links.

beshaya:
Looks like %PUBLIC_URL% does not resolve during compilation as I thought it would.


beshaya:
Now it works!

image

catherio (approved):
LGTM

PR #281, opened 2020-09-27
Clarify longer infection chains

### LINKED ISSUE #217: Q&A: Clarify "transmission chains" explanation
image

The first time I read this, I accidentally misinterpreted it to mean "I don't need to think through longer transmission chains for people I hang out with."

It would be useful to add a sentence along the lines of... What we are saying here only applies to longer transmission chains _for people you live with_. If you want to know how we model longer transmission chains, see (XYZ answer in the Q&A or the Research Sources).

👍

blanchardjeremy:
Fixes #217


src/paper/13-q-and-a.ts
```diff
@@ -22,7 +22,7 @@ In the case of a long indoor hangout, we suggest just to cap the total Activity
 
 ### If you live with multiple people, do you account for the additional risk of a longer infection chain?
 
-No. We’ll explain what this means and why we don’t account for it.
+No. We’ll explain what this means and why we don’t account for longer infections _with housemates_. _(Note: Longer infection chains are still important to account for when interacting with anyone who is not a housemate.)_
```
oremanj, line 25:
```suggestion
No. We’ll explain what this means and why we don’t account for it in the specific case of a single household.
```

src/paper/13-q-and-a.ts
```diff
@@ -33,6 +33,8 @@ There’s a \`30%\` chance that I get sick directly from Alice. There’s a \`70
 
 We currently don’t model this effect, and we think that’s pretty safe if (and only if) you and your household members are conscientious about monitoring for symptoms and isolating from one another at the first sign that someone in the house is feeling unwell. Such isolation should include some reduction in contact between the household members that still feel fine, since one or more of them might have become infected by the unwell person before their symptoms appeared.
```
oremanj, line 34: And then maybe add a footnote after "feeling unwell" here, to the effect of: "You probably can't do this anywhere near as reliably with someone you don't live with, which is why the discussion here is specific to housemates."

src/paper/13-q-and-a.ts
```diff
@@ -22,7 +22,7 @@ In the case of a long indoor hangout, we suggest just to cap the total Activity
 
 ### If you live with multiple people, do you account for the additional risk of a longer infection chain?
```
oremanj, line 23: Maybe add "within the household" at the end of this question to make it extra clear

oremanj (commented):
I think this needs a little more explanation about why we're distinguishing between housemates and non-housemates, and that some rewording of the question could reduce the need to insist quite so strongly on the distinction (because it will be clearer from the get-go). Specific suggestions inline.

blanchardjeremy:
Thanks for the suggestions. Let me know what you think of the latest commit.

blanchardjeremy:
@oremanj this one also looks good to go if you are ready to merge it.

PR #282, opened 2020-09-27
Add dropdown to white paper navigation. 

### LINKED ISSUE #248: UI: Add drop down menu under whitepaper
Minor suggestion: It'd be nice to navigate to the different whitepaper sections from a dropdown on the main menu. When I'm trying to get between sections, I have to keep returning to the main white paper page.

Alternatively, you could place the submenu on the whitepaper pages only as a sidebar (or floating sidebar).

blanchardjeremy:
Fixes #248

I didn't get around to adding active item highlighting. If folks feel that's a blocking issue, I can dig into it more.


src/App.tsx
```diff
@@ -61,13 +62,19 @@ export const App = (): React.ReactElement => {
...
                 <Nav.Item>
-                  <NavLink
-                    to="/paper"
-                    className="nav-link"
-                    activeClassName="active"
-                  >
-                    White Paper
-                  </NavLink>
+                  <NavDropdown title="White Paper" id="basic-nav-dropdown">
+                    <NavDropdown.Item href="/paper">
```
oremanj, line 66: I wonder if active item marking is as simple as removing the href here and adding a `<NavLink>` child? Not a blocker if it's not, but would be nice to have.

src/App.tsx
```diff
@@ -61,13 +62,19 @@ export const App = (): React.ReactElement => {
...
+                      Table of Contents
+                    </NavDropdown.Item>
+                    {Object.keys(pages).map((pageId, pageIndex) => (
+                      <NavDropdown.Item
+                        href={`/paper/${pageId}`}
+                        key={pageIndex}
+                      >
+                        {pages[pageId].shortTitle}
```
oremanj, line 74: Can we add the numbers here to more obviously distinguish the TOC from the subpages? Like `1: Introduction` or similar.

oremanj:
This is great, thanks! Couple of suggestions, feel free to punt if they're nontrivial.

src/App.tsx
blanchardjeremy: Good idea! On it!

src/App.tsx
blanchardjeremy: I tried out that suggestion and I couldn't get it working. I am brand new to React, so that's that main issue. ;) Maybe someone else can take a stab at it?

src/App.tsx
blanchardjeremy: Done.

blanchardjeremy:
Okay. I've completed what I can do with ease. If someone else wants to take a stab at the active menu highlighting, I support that. Otherwise we can merge it and come back to that later.

blanchardjeremy:
@oremanj This is no longer ready to be deployed. 

I have a version that is closer to functional. But it uses `window.location` which results in the highlighting not working if you switch to non-dropdown pages. (The highlighting stays on.) I'm brand new to React, so I have no idea out how to get the Router location available where I need it.

Options:

1. Either we publish it without highlighting ( 9e73ad7 )
2. Or someone helps me build on my latest commit ( eb946d6 ) to use proper URL matching.

Let me know your thoughts!

blanchardjeremy:
I just removed the not-yet-functional highlighting solution I came up with so this could actually be merged in with one click.

I can come back to proper highlighting later. Or if someone knows how to do it, please submit a separate PR.

PR #264, opened 2020-09-16
Add develop branch to workflow

beshaya:
This causes CI to run on pull requests into `develop` in case we want to use develop as a staging branch.

PR #288, opened 2020-09-30
Ben/voting

beshaya:


src/data/prepopulated.ts
```diff
@@ -182,6 +182,19 @@ export const prepopulated: {
...
   },
 
+  'Voting in-person': {
+    riskProfile: 'average',
+    interaction: 'oneTime',
+    personCount: 2,
+
+    setting: 'indoor',
+    distance: 'sixFt',
+    duration: 10,
```
blanchardjeremy, line 192: Hmmm. I guess it depends on what kind of voting we're talking about. A ballot drop-off or filling out a whole ballot. What had you think 10 minutes?

.github/workflows/ci.yml
```diff
@@ -4,9 +4,9 @@ name: CI
 # events but only for the master branch
 on:
   push:
-    branches: [ main ]
+    branches: [ main, develop ]
   pull_request:
-    branches: [ main ]
+    branches: [ main, develop ]
```
blanchardjeremy, line 9: Your #264 PR seems to be mixed in with this one. Do you wanna separate them out?

blanchardjeremy (changes requested):
Looks good overall. Take a look at the two comments and let me know what you think.

src/data/prepopulated.ts
oremanj: I think 10min is a reasonable estimate for how long it takes to sign the registry, get your ballot, and fill it out. Plus whatever part of the line is indoors, if there's a line.

beshaya: I just copied what Josh said in the bug. Do you think the numbers should be different? I haven't voted in person in a loooong time.

I'd be happy to add a mail-in dropoff scenario as well.

blanchardjeremy: I don't know if I've ever voted in-person, so I'm not a reliable source, either. Let's keep it at 10, then!

Seems like a ballot drop-off wouldn't be all that useful, given that this is already so few points.

oremanj: I think mail-in dropoff is basically free.

.github/workflows/ci.yml
blanchardjeremy: I went ahead and removed the devleop branch stuff myself.

blanchardjeremy:
@oremanj I say this is good to go!

blanchardjeremy (approved):
Good to go!

PR #293, opened 2020-10-04
Add husky lint-staged as a precommit script to prevent future lint issues.

blanchardjeremy:
There have been a bunch of lint issues lately. So I figured we could just add it as a pre-commit hook. I'd love someone to check my work who is more familiar with `lint-staged` and `husky`.


beshaya (approved):
This is magic and I love it. Thanks, Jeremy!

PR #296, opened 2020-10-05
Make risk modifiers visible. 

### LINKED ISSUE #229: Make relative risk visible when modifying inputs
When modifying inputs in the calculator, it would be useful to see the relative risk of each option in the dropdown during editing. This will save time and help make the logic behind the calculator more intuitive to users over time as they make more calculations.

At a minimum, it would be useful to sort options form lowest risk to highest risk and specifying this. Ex:

Surgical mask or mask with PM2.5 filter insert (lowest risk)
Cotton mask, bandanna, or buff
No mask (highest risk)

At best, it would be useful to see something like the following This would especially be helpful for cases where the relative risk may not be obvious. Ex:

Surgical mask or mask with PM2.5 filter insert (currently selected)
Cotton mask, bandanna, or buff (same risk)
No mask (5x riskier)

Long term, it may be worth considering replacing the dropdowns with sliders or another component that helps visualize relative positions on the risk spectrum.

blanchardjeremy:
I whipped this up.

I think it's worth publishing as-is just to have something out in the wild to make it more usable.

AND I think it would be better to do something like "1/20th the risk" instead of "0.05x risk".

So I'm gonna play with that.

Fixes #229

### DIFF
--- a/src/components/calculator/SelectControl.tsx
+++ b/src/components/calculator/SelectControl.tsx
@@ -7,7 +7,7 @@ import { FormValue } from 'data/data'
 
 function showRiskMultiplier(multiplier: number): string {
   if (multiplier === 1) {
-    return 'default risk'
+    return 'baseline risk'
   } else if (multiplier > 0 && multiplier < 1) {
     return `${num2fraction(multiplier)}th the risk`
   } else {
--- a/src/components/calculator/SelectControl.tsx
+++ b/src/components/calculator/SelectControl.tsx
@@ -1,9 +1,9 @@
+import num2fraction from 'num2fraction'
 import React from 'react'
 import { Badge, OverlayTrigger } from 'react-bootstrap'
 
 import { CalculatorData } from 'data/calculate'
 import { FormValue } from 'data/data'
-const fracty = require('fracty');
 
 function showRiskMultiplier(multiplier: number): string {
   // console.log(multiplier)
@@ -12,7 +12,7 @@ function showRiskMultiplier(multiplier: number): string {
   if (multiplier === 1) {
     return '(default risk)'
   } else if (multiplier > 0 && multiplier < 1) {
-    return `(${fracty(multiplier)}th the risk)`
+    return `(${num2fraction(multiplier)}th the risk)`
   } else {
     return `(${multiplier}x the risk)`
   }
--- a/src/components/calculator/SelectControl.tsx
+++ b/src/components/calculator/SelectControl.tsx
@@ -6,9 +6,6 @@ import { CalculatorData } from 'data/calculate'
 import { FormValue } from 'data/data'
 
 function showRiskMultiplier(multiplier: number): string {
-  // console.log(multiplier)
-  // console.log(multiplier < 0)
-  // debugger
   if (multiplier === 1) {
     return 'default risk'
   } else if (multiplier > 0 && multiplier < 1) {
--- a/src/components/calculator/SelectControl.tsx
+++ b/src/components/calculator/SelectControl.tsx
@@ -5,11 +5,21 @@ import { Badge, OverlayTrigger } from 'react-bootstrap'
 import { CalculatorData } from 'data/calculate'
 import { FormValue } from 'data/data'
 
+function getSuffix(frac: string): string {
+  const denom = frac.split('/')[1]
+  if (denom === '2') return ''
+  else if (denom === '3') return 'rd'
+  else return 'th'
+}
+
 function showRiskMultiplier(multiplier: number): string {
   if (multiplier === 1) {
     return 'baseline risk'
   } else if (multiplier > 0 && multiplier < 1) {
-    return `${num2fraction(multiplier)}th the risk`
+    // Fraction format: "1/5th the risk"
+    console.log(`test: ${num2fraction(0.3)}`)
+    const frac = num2fraction(multiplier)
+    return `${frac}${getSuffix(frac)} the risk`
   } else {
     return `${multiplier}x the risk`
   }
--- a/src/components/calculator/SelectControl.tsx
+++ b/src/components/calculator/SelectControl.tsx
@@ -25,7 +25,7 @@ export const GenericSelectControl: React.FunctionComponent<{
   value: string | number
   label?: string
   popover?: JSX.Element
-  hiderisk?: boolean
+  hideRisk?: boolean
 }> = (props) => (
   <div className="form-group">
     {props.label && (
@@ -52,7 +52,7 @@ export const GenericSelectControl: React.FunctionComponent<{
       {Object.keys(props.source).map((value, index) => (
         <option key={index} value={value}>
           {props.source[value].label}{' '}
-          {!props.hiderisk &&
+          {props.hideRisk !== true &&
             `[${showRiskMultiplier(props.source[value].multiplier)}]`}
         </option>
       ))}
@@ -68,7 +68,7 @@ export const SelectControl: React.FunctionComponent<{
   source: { [key: string]: FormValue }
   label?: string
   popover?: JSX.Element
-  hiderisk?: boolean
+  hideRisk?: boolean
 }> = (props) => (
   <GenericSelectControl
     id={props.id}
@@ -77,7 +77,7 @@ export const SelectControl: React.FunctionComponent<{
     value={props.data[props.id] || ''}
     label={props.label}
     popover={props.popover}
-    hiderisk={props.hiderisk}
+    hideRisk={props.hideRisk}
   />
 )
 
--- a/src/components/calculator/SelectControl.tsx
+++ b/src/components/calculator/SelectControl.tsx
@@ -11,6 +11,7 @@ export const GenericSelectControl: React.FunctionComponent<{
   value: string | number
   label?: string
   popover?: JSX.Element
+  hiderisk?: boolean
 }> = (props) => (
   <div className="form-group">
     {props.label && (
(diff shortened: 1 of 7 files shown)


blanchardjeremy:
image

Okay I went for it and added fractional modifiers. Let me know what y'all think.

It has a lint error. I'm just learning ES6 and React, so I need some help to figure out how to switch `fracty` from a `require()` statement to an import. Help?

src/components/calculator/SelectControl.tsx
```diff
@@ -3,6 +3,20 @@ import { Badge, OverlayTrigger } from 'react-bootstrap'
...
+function showRiskMultiplier(multiplier: number): string {
+  // console.log(multiplier)
+  // console.log(multiplier < 0)
+  // debugger
+  if (multiplier === 1) {
+    return '(default risk)'
+  } else if (multiplier > 0 && multiplier < 1) {
+    return `(${fracty(multiplier)}th the risk)`
+  } else {
+    return `(${multiplier}x the risk)`
```
blanchardjeremy, line 17: Perhaps we should put the risk in square brackets instead of parentheses?

beshaya:
Thanks for taking care of this. Two things I'd like to see cleaned up:

1) The type of interaction dropdown doesn't display the multipliers (but only if it was set to "select one" when clicked?!)
image

2) The fractions look kind of icky for odd numbers (specifically partner):
image

I propose rounding our calculator numbers to friendly fractions (i.e. partner risk from 48%->1/2, housemate 30%->1/3). I think the rest are clean already. I'll bring this up in slack as well.

src/components/calculator/SelectControl.tsx
```diff
@@ -1,16 +1,31 @@
+import num2fraction from 'num2fraction'
 import React from 'react'
 import { Badge, OverlayTrigger } from 'react-bootstrap'
 
 import { CalculatorData } from 'data/calculate'
 import { FormValue } from 'data/data'
 
+function showRiskMultiplier(multiplier: number): string {
+  // console.log(multiplier)
```
beshaya, line 9: Delete comments

src/components/calculator/SelectControl.tsx
```diff
@@ -1,16 +1,31 @@
...
+  // console.log(multiplier < 0)
+  // debugger
+  if (multiplier === 1) {
+    return 'default risk'
+  } else if (multiplier > 0 && multiplier < 1) {
+    return `${num2fraction(multiplier)}th the risk`
```
beshaya, line 15: Could you add custom text for the "Type of Interaction" dropdown? This text doesn't reflect what that dropdown does. It should be:

One time [6% per hour]
Housemate [30% each week]
Spouse / Partner [48% each week]

blanchardjeremy:
> 1. The type of interaction dropdown doesn't display the multipliers (but only if it was set to "select one" when clicked?!)

I intentionally hid it on this dropdown. Because for "type of activity" it isn't actually a modifier from a baseline in the same way. It serves as a cap for the upper limit of risk of the activity. I don't see showing ".06x risk" making sense for one-time. It is a very different thing than masks.

> 2. The fractions look kind of icky for odd numbers (specifically partner):
Oh I just didn't want it to show at all. So let me fix that!

src/components/calculator/SelectControl.tsx
```diff
@@ -1,16 +1,31 @@
...
 
 import { CalculatorData } from 'data/calculate'
 import { FormValue } from 'data/data'
 
+function showRiskMultiplier(multiplier: number): string {
+  // console.log(multiplier)
+  // console.log(multiplier < 0)
+  // debugger
+  if (multiplier === 1) {
+    return 'default risk'
```
beshaya, line 13: I think I prefer [1x risk]. "Default" and "1x" are both showing that this is an implementation detail, so I'd go with the more consistent format.

Likewise, maybe do `num2fraction x risk`

_Additionally_ maybe lets do "transmission" or "chance of transmission" instead of "risk" as someone pointed out that we have used "risk" to mean way too many things.

src/components/calculator/SelectControl.tsx
```diff
@@ -36,7 +51,9 @@ export const GenericSelectControl: React.FunctionComponent<{
       <option value="">Select one...</option>
       {Object.keys(props.source).map((value, index) => (
         <option key={index} value={value}>
-          {props.source[value].label}
+          {props.source[value].label}{' '}
+          {!props.hiderisk &&
+            `[${showRiskMultiplier(props.source[value].multiplier)}]`}
```
beshaya, line 56: Can we throw this in an `<i>` tag for readability?

src/components/calculator/SelectControl.tsx
blanchardjeremy: Haha. Oops! fixing now.

src/components/calculator/SelectControl.tsx
blanchardjeremy: You can't use other HTML tags inside an <option> tag. Which is why I opened #298

src/components/calculator/SelectControl.tsx
blanchardjeremy: Smart. I like it. I'll just hard code that in.

src/components/calculator/SelectControl.tsx
blanchardjeremy: Regarding 1x risk: Definitely is an implementation detail. I don't know that "1x risk" means anything to a lot of people. It feels confusing to me if I put myself in the seat of a user. That said, I think "baseline risk" might be better than "default risk." Whatcha think?

> Likewise, maybe do num2fraction x risk

Do you mean doing "1/20 x risk" instead of "1/20th the risk"

> "chance of transmission" 

Hmmm. Makes the select boxes really long, which is challenging on mobile. Maybe we could switch to this once we implement better boxes (see #298).

blanchardjeremy:
@beshaya check out my updates.

src/components/calculator/SelectControl.tsx
beshaya: I think it's more cognitive load to see descriptors with different formats. Parallel structure!

I'm ok with "risk" but want to challenge us to find better but concise language.

beshaya:
> > 1. The type of interaction dropdown doesn't display the multipliers (but only if it was set to "select one" when clicked?!)
> 
> I intentionally hid it on this dropdown. Because for "type of activity" it isn't actually a modifier from a baseline in the same way. It serves as a cap for the upper limit of risk of the activity. I don't see showing ".06x risk" making sense for one-time. It is a very different thing than masks.

Fine to merge with it hiding. I think it would be *REALLY* good to explain what this dropdown does, (otherwise it is by far the most "magic" dropdown) but it can be in a follow-up.

src/data/data.ts
```diff
@@ -7,19 +9,27 @@ const formValue = function (label: string, multiplier: number): FormValue {
...
-    label: 'One-time interaction [6% chance of transmission per hour]',
-    multiplier: 0.06,
+    label: `One-time interaction [${fixedPointPrecisionPercent(
+      oneTimeMult,
+    )} chance of transmission per hour]`,
+    multiplier: oneTimeMult,
   },
   repeated: {
-    label: 'Household member [30% chance of transmission per week]',
+    label: `Household member [${fixedPointPrecisionPercent(
```
beshaya, line 23: Oooh, I don't know if it's a good idea to have the calculator use a different value from the label. I'm worried its going to confuse someone when the output results aren't actually doubling.

beshaya: I think either expose the 48% or change the value that the calculator uses to 50%.

beshaya (changes requested):

src/components/calculator/SelectControl.tsx
blanchardjeremy: @beshaya Can you clarify exactly what you'd have them say, then?

blanchardjeremy: Agreed about finding better language!

src/data/data.ts
blanchardjeremy: Oh I see. Well we already have this problem to some degree because the explanation text below has been showing 50% for a while: 
image

blanchardjeremy:
![image](https://user-images.githubusercontent.com/32290/95119873-86088880-0701-11eb-84ad-04d13d521daf.🤦 Also, I just caught a major bug!

Need to disable it for that dropdown too! 🤦‍♂️

src/data/data.ts
blanchardjeremy: I'm happy to go either way. What's currently is in this PR is consistent with that explanation text. I think we could make the change you're suggesting in this PR or in a separate thread. I'll follow your lead.

src/components/calculator/SelectControl.tsx
blanchardjeremy: Also, I was following @mattbell's suggested format from the original issue when using "default risk" vs "5x risk". I'm curious if he wants to weigh in on this.

src/components/calculator/SelectControl.tsx
blanchardjeremy: Let me know if you have any other ideas.

src/data/data.ts
beshaya: Yeah, I see that this is all over the place. I'll bring it up in a different bug.

src/components/calculator/SelectControl.tsx
beshaya: I'm ok punting this. <i> would have been nice if it was easy.

beshaya (approved):
LGTM

PR #297, opened 2020-10-05
Add "all on one page" to dropdown

catherio:
Now that there's a dropdown, and people mostly won't use the TOC page, it'll be harder for people to discover that there's an "all in one page" option. Adding it to the menu.


blanchardjeremy (approved):
Smart. Looks good.

beshaya (approved):
Nice!

PR #300, opened 2020-10-05
Use same header styles across the entire site

blanchardjeremy:
My two cents: it would be better if we used the same header styles across the entire site (both the white paper and the other pages).

Thoughts?


beshaya:
@dominic for styling review.

dominic (approved):

dominic:
LGTM 👍

PR #305, opened 2020-10-06
Ben/cuddling tweaks

beshaya:
Fixes: 265

### DIFF
--- a/src/components/calculator/ActivityRiskControls.tsx
+++ b/src/components/calculator/ActivityRiskControls.tsx
@@ -81,10 +81,9 @@ export const ActivityRiskControls: React.FunctionComponent<{
       {data.setting === 'outdoor' &&
       ['close', 'intimate'].includes(data.distance) ? (
         <div className="warning">
-          Due to the very close distances and exchange of bodily fluid, we are
-          not confident that being outdoors reduces the risk in a substantial
-          way. Thus, we are not providing any bonus for being outdoors when
-          intimate.
+          Due to very close distances, we are not confident that being outdoors
+          reduces the risk in a substantial way. Thus, we are not providing any
+          bonus for being outdoors when intimate.
         </div>
       ) : null}
       <div className="form-group">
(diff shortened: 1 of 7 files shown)


beshaya:
image

src/components/calculator/ActivityRiskControls.tsx
```diff
@@ -64,19 +64,27 @@ export const ActivityRiskControls: React.FunctionComponent<{
...
         setter={setter}
-        source={Distance}
+        source={Setting}
       />
+      {data.setting === 'outdoor' && data.distance === 'intimate' ? (
+        <div className="warning">
+          Due to the very close distances and exchange of bodily fluid, we are
+          not confident that being outdoors reduces the risk in a substantial
+          way. Thus, we are not providing any bonus for being outdoors when
+          intimate.
```
blanchardjeremy, line 85: Since this is doing double duty as "cuddling" and "kissing/having sex" I wonder if it should say something like: 

> Due to the very close distances and **possible** exchange of bodily fluids, we are...

src/data/data.ts
```diff
@@ -25,7 +25,7 @@ export const Setting: { [key: string]: FormValue } = {
   outdoor: { label: 'Outdoor', multiplier: 0.05 },
 }
 export const Distance: { [key: string]: FormValue } = {
-  intimate: formValue('Cuddling / intimate', 2),
+  intimate: formValue('Intimate', 2),
```
blanchardjeremy, line 28: I'm going to leave a comment on this in the issue.

blanchardjeremy (changes requested):
I'd suggest this change.

src/components/calculator/ActivityRiskControls.tsx
beshaya: Good catch. done.

blanchardjeremy (approved):
👍👍👍

PR #313, opened 2020-10-07
Move setQuery to button click

### LINKED ISSUE #307: UX Redesign: Move "share by link" features to a share button
Building on issue #193 and PR #291: 

We should not automatically add all the query variables to the URL every time a user changes something. Instead, we should have a "share this scenario by link" button that copies a link to the clipboard. 

Tagging @cameronmarlow who may be interested in helping.

cameronmarlow:
Add a button to the Calculator when points exists and trigger a query update when the button is pressed (as opposed to when the state is changed).

This does not trigger any share event per se, it still requires the reader to press "share" on their device. Unfortunately the [Web Share API](https://caniuse.com/web-share) is not in broad adoption so anything more direct will require a bespoke interface.

* Add another action button for updating the query
* Trigger query update on button press

Closes #307


cameronmarlow:
We could use [react-share](https://github.com/nygardk/react-share) in some form to replace the share button once the url has been updated or simply put the buttons in the UI and use the `beforeOnClick` in the share buttons to update the URL for consistency.

beshaya:
I think it would be clearer to show the share-able url in a text box (maybe with a copy button) over putting it in the url of the browser. The current behavior isn't easily discoverable (you have to know to look for the new url). Especially on mobile, where you don't actually see the URL unless you click the bar.

 @mattbell @blanchardjeremy what do you think?

blanchardjeremy:
@beshaya Thanks for having your eye on the mobile experience. I think we could do what you're proposing. Or we could have it say "copy link to clipboard" and then display a temporary notice saying "link copied to clipboard" .... This matches how most mobile apps would do it. Thoughts?

cameronmarlow:
See the most recent version - I added a share form with a copy-to-clipboard function.

You don't currently have bootstrap alerts setup anywhere, correct?

cameronmarlow:
Put another way - let me know how you would like me to notify the user that the URL has been copied. A few options:

* Save copied state and change the form to "Copied" when it has been copied, set copied to false when the form changes
* Add Bootstrap alerts with a simple "Copied to clipboard." I could also add an alert for the form saving while I'm at it.

Let me know what you prefer @beshaya or @blanchardjeremy

beshaya:
I think an alert would be most similar to what I see elsewhere on the web.
Adding an alert for form saving would be great as well.

cameronmarlow:
@beshaya take a look the the simple alerts in the most recent commit

blanchardjeremy:
Looking good so far

I'd still like to suggest that we never reveal the whole URL to people. It takes up space and doesn't seem that helpful. Since the only thing folks need is the "copy to clipboard" button anyway. 

I'd also like to suggest we add a "copy" icon to this button. 

Thoughts?

blanchardjeremy:
6916454B-6EF6-4D36-8E06-B264C21EA207


Also could we get some space between the buttons?

cameronmarlow:
I'm just following the design that was currently there for the form save, but happy to spruce it up a bit.

Screen Shot 2020-10-08 at 1 12 00 PM

cameronmarlow:
As for getting rid of the URL, that's a convention I have seen quite a bit (show the URL and have a button to copy it). If we don't want to show the URL, we have a couple of options:

1. Have the share button directly copy to the clipboard with the appropriate alert
2. Put the existing buttons on a separate line as they are now but without the input box
3. Another UI where we expose share links for other services in addition to copy-to-clipboard, e.g. twitter, fb

blanchardjeremy:
Oh I think having the two buttons next to each other (if we're keeping this design) makes sense. I meant vertical padding between this input group and the two other buttons.

As far as the URL pattern, yeah I definitely see the format you're describing some places. But I'd say the "copy straight to clipboard" format is more common. And it's a better mobile experience (and 50% of our users are on mobile).

Youtube:
image

Github:
image

Spotify:
image

To be fair, all three of those are menus, not buttons. And there are plenty of counterexamples.

If someone else wants to weigh in in favor of the current format, I'm down. My vote is one button with a notice.

blanchardjeremy (changes requested):

blanchardjeremy:
@cameronmarlow See #371. I went ahead with the variation I prefer over there. Curious to hear everyone's thoughts.

beshaya:
Landed in the ux-redesign branch as #375

Issue #310, opened 2020-10-07
Differentiate more clearly between "weekly" and "each time" types


blanchardjeremy:
I sat down with my housemate who is a UX designer to look at @mattbell's latest mockups (see #294).

One of our biggest takeaways is that there are basically two very different "modes" for the calculator: 1) a result that is a "each time" score. And 2) a result that is a "per week" score (partner/housemate). 

I think there's more for us to do to indicate very clearly to the user when the switch: "The score you were just seeing means something very very different than the result you are now seeing."

Having the % of weekly budget section helps some.

I don't have a good solution to this yet. I wonder if there's some kind of visual icon we can add to indicate "score per time you do it" vs "score for the week." 

I wanted to raise it so we could bring our creative ideas to the table.

mattbell:
I agree with the general concern.  I think adding "per week" or "each time" to the end of the microcovid result number is a good short term solution.  If you can convey that with icons while taking up less space, great. :-)

In the medium term I think we probably need a better way of asking the follow-on questions for housemates / intimate partners so people understand how to select risk profiles better.

catherio:
calculator used to say this but there was a regression! https://github.com/microcovid/microcovid/issues/221#issuecomment-691799124

am 👍👍👍👍 to fixing this!!!

blanchardjeremy:
@catherio - as a short term fix, here's a PR: #315. 

Keeping this issue open so we can look at a longer-term fix.

PR #315, opened 2020-10-07
Fix 'each time' after points display. 

blanchardjeremy:
Short term solution to  #310


beshaya:
Do you think that this is unclear without the extra words? We're already so darn wordy...

beshaya (approved):

PR #318, opened 2020-10-08
Don't pass undefined variables from QueryParams.

beshaya:
Also does stricter sanity checking within PrevalenceControls, such that this
it will be resilient unsanitzed data in the future.

Fixes: 317


blanchardjeremy (approved):
looks good.

Issue #60, opened 2020-08-27
It is hard to see how to enter specific prevalence data once you have selected a location already


david-mears:
On the calculator, the initial state allows you to enter your own prevalence data, e.g. if you know about your local area and want to use that more accurate data.

However, if you select a location, it is very difficult to see how to then revert your decision, and input your own data. Originally, when making this issue, I believed it was impossible to revert your decision. I had been going to complain that because the form selection is persisted between page reloads, you cannot get around this by reloading the page.

However, I have since found that clicking on 'Select location or enter data' in the first of the two location drop-downs allows me to enter my own data.

It's not just me who has found it difficult to find the way to revert the decision. I shared the site with my partner (while telling her not to share it on further) yesterday, and she commented on this (unsolicited):

> it's a shame microcovids doesn't let you have more granularity than 'England'. It would be good if i could enter cases in Hackney or London  + population, because i'm guessing the risk that the calculator outputs will be 'diluted' by all the less-risky parts of england

> i get that they can't have access themselves to data for everywhere in the world, but it looks like i ought to be able to adjust the pop and case numbers myself, but I can't.

I suggest adding a button saying 'Fill in own data' which has the same effect as clicking 'Select location or enter data'. I most likely will not have time to make this button myself before launch.

beshaya:
Good point, David. We should add a button to clear the location data, but we honestly might not get to that for a while. If you wanted to send a PR for that over, we'd be happy to take it.

In the meanwhile, here are two workarounds:
* hit "reset"
* set the location to "Select location or enter data..."

PR #326, opened 2020-10-09
Add button to fix issue 60

david-mears-dfe:
Fixes https://github.com/microcovid/microcovid/issues/60

## Description

(Please note that I am a Ruby developer, not a React developer, so there may be whoopsies.)

This displays a `or enter data` button if no top location has been set. The button selects `Select a location or enter data...` on the top location drop-down. The button does not exist when `Select a location or enter data...` is selected.

'Step 1 - Choose a location' has been edited to say 'Step 1 - Location Risk' (to match the Person Risk and Activity Risk sections, and to avoid implying that the only method of getting the location risk is by choosing a location from the drop-down.)

## Screenshots

Screenshot 2020-10-09 at 14 53 27
Screenshot 2020-10-09 at 14 53 36

### DIFF
--- a/src/components/calculator/PrevalenceControls.tsx
+++ b/src/components/calculator/PrevalenceControls.tsx
@@ -70,10 +70,8 @@ const PrevalenceField: React.FunctionComponent<{
       onChange={(e) => {
         if (isNumber(max) || isNumber(min)) {
           let newValue = Number.parseFloat(e.target.value)
-          console.log(newValue)
           newValue = isNumber(max) && newValue > max ? max : newValue
           newValue = isNumber(min) && newValue < min ? min : newValue
-          console.log(newValue)
           setter(newValue.toString())
         } else {
           setter(e.target.value)
--- a/src/components/calculator/PrevalenceControls.tsx
+++ b/src/components/calculator/PrevalenceControls.tsx
@@ -108,7 +108,7 @@ export const PrevalenceResult = (props: {
   data: CalculatorData
 }): React.ReactElement => {
   return (
-    <Card className="prevelance-result">
+    <Card className="prevalence-result">
       <Card.Body>
         <div>
           <Trans>calculator.prevalence.reported_prevalence</Trans>:{' '}
@@ -251,7 +251,7 @@ export const PrevalenceControls: React.FunctionComponent<{
       )}
 
       <ControlledExpandable
-        id="prevelance-details"
+        id="prevalence-details"
         header={t('calculator.prevalence.details_header')}
         headerClassName={isManualEntryCurrently ? 'd-none' : ''}
         open={detailsOpen}
--- a/src/components/calculator/PrevalenceControls.tsx
+++ b/src/components/calculator/PrevalenceControls.tsx
@@ -148,10 +148,6 @@ export const PrevalenceControls: React.FunctionComponent<{
     }
   }
 
-  const handleEnterDataButtonOnClick = () => {
-    setLocationData(TOP_LOCATION_MANUAL_ENTRY, '')
-  }
-
   const setLocationData = (topLocation: string, subLocation: string) => {
     if (isManualEntry(topLocation)) {
       setDetailsOpen(true)
@@ -162,6 +158,14 @@ export const PrevalenceControls: React.FunctionComponent<{
       topLocation,
       subLocation,
     })
+  }
+
+  const handleEnterDataButtonOnClick = () => {
+    setLocationData(TOP_LOCATION_MANUAL_ENTRY, '')
+  }
+
+  const handleSelectLocationButtonOnClick = () => {
+    setLocationData('', '')
   }
 
   // If a stored location exists, load latest data for that location.
@@ -189,6 +193,11 @@ export const PrevalenceControls: React.FunctionComponent<{
     Locations[data.topLocation].subdivisions.length > 1
 
   const locationSet = isTopLocation(data.topLocation)
+
+  const showEnterDataButton =
+    locationSet ||
+    !isFilled(data.topLocation) ||
+    data.topLocation !== TOP_LOCATION_MANUAL_ENTRY
 
   const [detailsOpen, setDetailsOpen] = useState(
     false || isManualEntry(data.topLocation),
@@ -226,15 +235,6 @@ export const PrevalenceControls: React.FunctionComponent<{
             </optgroup>
           ))}
         </select>
-        {!locationSet ? null : (
-          <button
-            id="enterData"
-            className="btn btn-secondary"
-            onClick={handleEnterDataButtonOnClick}
-          >
-            {t('calculator.select_location_enter_manually')}
-          </button>
-        )}
       </div>
       {!showSubLocation ? null : (
         <div className="form-group">
@@ -260,7 +260,23 @@ export const PrevalenceControls: React.FunctionComponent<{
           </select>
         </div>
       )}
-
+      {!showEnterDataButton ? (
+        <button
+          id="switchBetweenManualDataAndLocationSelection"
+          className="btn btn-secondary"
+          onClick={handleSelectLocationButtonOnClick}
+        >
+          {t('calculator.switch_button.select_location')}
+        </button>
+      ) : (
+        <button
+          id="switchBetweenManualDataAndLocationSelection"
+          className="btn btn-secondary"
+          onClick={handleEnterDataButtonOnClick}
+        >
+          {t('calculator.switch_button.enter_data_manually')}
+        </button>
+      )}
       <ControlledExpandable
         id="prevalence-details"
         header={t('calculator.prevalence.details_header')}
--- a/src/components/calculator/PrevalenceControls.tsx
+++ b/src/components/calculator/PrevalenceControls.tsx
@@ -148,6 +148,10 @@ export const PrevalenceControls: React.FunctionComponent<{
     }
   }
 
+  const handleEnterDataButtonOnClick = () => {
+    setLocationData(TOP_LOCATION_MANUAL_ENTRY, '')
+  }
+
   const setLocationData = (topLocation: string, subLocation: string) => {
     if (isManualEntry(topLocation)) {
       setDetailsOpen(true)
@@ -222,6 +226,15 @@ export const PrevalenceControls: React.FunctionComponent<{
             </optgroup>
           ))}
         </select>
+        {!locationSet ? null : (
+          <button
+            id="enterData"
+            className="btn btn-secondary"
+            onClick={handleEnterDataButtonOnClick}
+          >
+            {t('calculator.select_location_enter_manually')}
+          </button>
+        )}
       </div>
       {!showSubLocation ? null : (
         <div className="form-group">
--- a/src/components/calculator/PrevalenceControls.tsx
+++ b/src/components/calculator/PrevalenceControls.tsx
@@ -194,23 +194,18 @@ export const PrevalenceControls: React.FunctionComponent<{
 
   const locationSet = isTopLocation(data.topLocation)
 
-  const showEnterDataButton =
-    locationSet ||
-    !isFilled(data.topLocation) ||
-    data.topLocation !== TOP_LOCATION_MANUAL_ENTRY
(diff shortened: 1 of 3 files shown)


david-mears-dfe:
@beshaya @dominic have you seen this?

beshaya:
The team is working on a major overhaul of the calculator UX, and we are including something similar. Check out the ux-redesign branch or the preview at https://ux-redesign--microcov.netlify.app/ for what's happening.

david-mears-dfe:
> The team is working on a major overhaul of the calculator UX, and we are including something similar. Check out the ux-redesign branch or the preview at https://ux-redesign--microcov.netlify.app/ for what's happening.

Thanks for the reply @beshaya. Would you welcome a similar PR to that branch?

david-mears-dfe:
(As I don't see an equivalent button on that preview url)

beshaya:
Yeah, you're right. I think the button is easier to see. @blanchardjeremy what do you think?

blanchardjeremy:
Hmmm. I agree that button is easier to see. This is a convo we've been having over the past few weeks. 

I opted for a drop-down menu item because I don't think that many users actually want to enter data manually. It's a fairly advanced operation. I imagine people would opt for country level data even if we don't have enough their particular region.

But I have no data to back up my guess! 😂 

@david-mears-dfe are you someone who's entering data manually? If so can you tell us more about your use case? And why you'd prefer a button over the drop-down menu item?

blanchardjeremy:
Oh I guess I should also ask: did you see that this feature does exist as a drop-down item? Maybe that wasn't already clear.

david-mears-dfe:
> Oh I guess I should also ask: did you see that this feature does exist as a drop-down item? Maybe that wasn't already clear.

It wasn't. There is a lot of detail about this described in the issue this PR fixes. https://github.com/microcovid/microcovid/issues/60.

> are you someone who's entering data manually? If so can you tell us more about your use case? And why you'd prefer a button over the drop-down menu item?

I'm entering data manually because averaging across England is not accurate enough. This variance makes a large difference to what activities are tolerable. From the UK government's point of view, there are 3 very different 'tiers' of lockdown across the country according to local prevalence. In Scotland they use 5 tiers.

I've now 'advanced' to using the spreadsheet, but in my initial forays into microcovid (a necessary precursor to using the spreadsheet?) I used the calculator.

blanchardjeremy:
@david-mears-dfe is this still something you're interested in submitting a PR for? If so, let us know.

david-mears-dfe:
Yes it is, I'm going to look at it today.

david-mears-dfe:
I can't see anything called MANUAL_DATA_KEY in the code, but my plan is to have the button simply select the drop-down option (as in this original PR).

david-mears-dfe:
Ok I've gone ahead and updated this PR.

Each change is divided into very atomic commits so that I can easily discard any changes you don't like.

Then I can squash the commits if you would like.

blanchardjeremy:
@david-mears-dfe @david-mears - looking good so far. 

Is the main use case you want to solve "how to switch back to manual location data"? I ask because I figured you would also want the button to be visible when nothing was selected in either drop-down. So that folks could find it more easily. 

Also I suggest we move the button below the second drop-down? I think that fits better visually and contextually. Let me know. 

Thanks for doing this!

david-mears-dfe:
Re-updated. I have updated the behaviour as you suggest and the position. I've also added a button to return to inputting a location which is visible in 'enter data' mode.

david-mears-dfe:
@blanchardjeremy Would you have time to take a look this week? I don't want this to get too far out of step with the latest master.

blanchardjeremy:
Thanks for following up. At this point, I'm not sure I'm the best person to see this one through.  I find the current layout confusing, but I don't have a suggestion for how to make it better.  I don't want to be a blocker on forward progress. So I'll ask if anyone else on the team might be willing to chime in. Maybe @beshaya

blanchardjeremy:
I propose we close this as there hasn't been any movement on it in a while.

david-mears-dfe:
I'm sending some feedback on the contribution process to @catherio .

blanchardjeremy:
@david-mears-dfe Hey. Thanks for sending along feedback in that way. If you feel comfortable, I'd invite your feedback here as well so we can attempt to clear it up directly. I'll say upfront: I'm sorry that this hasn't been brought to completion.  I can understand it being frustrating to have something you want to contribute and it not being seen through. Perhaps it was short-sighted of me to propose closing this rather than giving it a clear path forward. I'd be open to hearing what you want to share and I'm sure the rest of the team would appreciate it too. Let us know what you feel comfortable with. (This PR seems as a good a place as any for this conversation, IMO. Or I suppose we could move it over to [Discussions](https://github.com/microcovid/microcovid/discussions))

david-mears-dfe:
Thanks @blanchardjeremy. I sent the feedback to Catherine as a precaution because I couldn't be sure you'd like to have the feedback public. Here's what I wrote.

> This is some hopefully constructive criticism for the microcovids team coming from a place of frustration.

> I think it might be worth the microcovids team thinking about how to make it easier to contribute changes to the project (or it might not, e.g. if the pandemic is over soon).

> After having it confirmed that a PR would be welcomed on this issue, I submitted the PR, and updated it in response to feedback, believing I was making a noticeable usability improvement to your site. But then I had my work go to waste because of a lack of timely engagement. If it hadn't gone to waste, I can confidently say I would have continued to make and submit fixes based on your issue backlog (instead, I've chosen other EA projects to contribute to).

> In the end, I wasted a lot of time. This was frustrating for me but not necessarily any one person's fault. From your perspective, I can understand that it's easily possible that the overhead of dealing with my PRs would have outweighed their benefit to users. And it's easy to overestimate how much capacity a team will have for dealing with external PRs, and to overlook designing a good process for contributors.

> One idea for what I would have preferred is: your team (or an initiative-taker) decided on a particular person to 'own' the PR, who then is empowered to decide at any point that it's not worth continuing with it, and who is responsible for communicating that to the author, and for avoiding the failure mode of the work being left there until it's incompatible or stale.

> Going forward, I think you could do something like that, rather than having a de facto halfway house policy that wastes time on both sides and goodwill on the contributors' side; or maybe you don't need a policy per se but would just benefit from being more aware of this blindspot/failure mode.

I'm open to re-updating the branch once more, iff you want me to and iff it's unlikely to go stale again.

blanchardjeremy:
@david-mears-dfe I just had a chance to read your message. Thank you for taking the time to share this. I really appreciate the thoughtful and compassionate way you shared it! Makes it easy to engage with and reply to. 👍

Since I've been the main person tracking this and engaging with you, I think most of how this went rests with me. I hear you when you say it's frustrating to have a PR invited and then dropped the way this was dropped. Sorry for my part in that, in "punting" it to the team without making sure someone picked up the ball. I think your suggestion about having a clear point person for each PR makes a lot of sense. Seems like a way to support people to have a successful experience contributing. It's important to me (and the team) that folks are supported to make the contributions that make sense to make, and that people not waste their time on things that aren't likely to be accepted.

blanchardjeremy:
As far as this PR goes, if you'd like to still contribute this, I"m down to be the point person for it!  I think the desired behavior that would resolve my earlier confusion would basically be something like:

* A button below both of the drop-downs that says "Switch to manual data entry mode" 
* Remove "enter data manually" from the drop-down menu. --- I could be swayed that it makes sense to have it in both places. But my first thought is that it seems confusing to me. Better to have it in just one place, IMO. The trick with implementing this approach will be that that field still needs to contain the MANUAL_DATA key so that it works with shareable URLs. So you'd have to add a bit of code in a few places to set that key into the field (and URL) and pull it out of the URL on page load.
* When you click the bottom, both dropdowns are hidden and button text changes to "Switch to pre-populated location data" (or something like that)

Those are my ideas. I'm open to revisions. Let me know your thoughts!

david-mears-dfe:
Thanks @blanchardjeremy. I'll take a look at the weekend.

Reminded of:
image

david-mears-dfe:
@blanchardjeremy I've added the behaviour you describe. It was less complicated than expected to preserve the shareable url functionality.

blanchardjeremy:
This is looking really good to me!
Only thing that seems off is this: when you're on automatic location data mode, it no longer shows the extra info (like cases last week). If you compre the deploy preview to the current live site, you should see the difference.

src/components/calculator/PrevalenceControls.tsx
```diff
@@ -249,78 +257,98 @@ export const PrevalenceControls: React.FunctionComponent<{
...
-            unit="%"
-            setter={(_value) => null}
-            inputType="text"
-            isEditable={false}
-          />
+        {!isManualEntryCurrently ? (
+          <PrevalenceResult data={data} />
         ) : (
-          <PrevalenceField
-            id="positive-test-rate"
```
blanchardjeremy, line 311: I'd still like to see these fields show up even if automatic data location is being used:

image

blanchardjeremy (changes requested):

david-mears-dfe:
I fixed the issue you mention @blanchardjeremy .

While working on this I also noticed the following issue, both on this branch and on the live site:

When the location is set to 'Select location', and you click the Details drop down, the fields are editable. Is this intentional? I think it isn't, because if you fill the fields, it doesn't behave as if you are it manual data-entry mode. I think the fix will be to use `isManualDataCurrently` rather than `!locationSet` for the PrevalenceFields' isEditable attribute. I can submit this change in a separate PR as it's not related to this PR.

blanchardjeremy:
> While working on this I also noticed the following issue, both on this branch and on the live site:
> 
> When the location is set to 'Select location', and you click the Details drop down, the fields are editable. Is this intentional? I think it isn't, but I'm not sure. I think the fix will be to use `isManualDataCurrently` rather than `!locationSet` for the PrevalenceFields' isEditable attribute. I can submit this change in a separate PR as it's not related to this PR.

Oh good catch. Yes, happy to have a PR on that issue as well. 👍

blanchardjeremy:
And one last fix. 

image


After the user types in the manual data, I think they should see the actual prevalence and adjusted prevalence numbers still. The live site has the functionality.

david-mears-dfe:
> And one last fix.
> 
> After the user types in the manual data, I think they should see the actual prevalence and adjusted prevalence numbers still. The live site has the functionality.

Are you sure? I can't see this behaviour on microcovid.org. But it is simple to add. The change would be from:
```
{!isManualEntryCurrently && <PrevalenceResult data={data} />}
```
to:
```
<PrevalenceResult data={data} />
```

@blanchardjeremy 

Screenshot 2021-01-15 at 18 35 01

src/locales/en.json
```diff
@@ -109,9 +109,12 @@
     "location_subprompt_US-LA": "Entire state, or select parish...",
     "location_subprompt_US-AK": "Entire state, or select borough...",
     "location_subprompt_US": "Entire state, or select county...",
-    "location_selector_header": "Step 1: Enter your location",
+    "location_selector_header": "Step 1: Location Risk",
```
blanchardjeremy, line 112: For this one, I'd prefer to opt for "Enter your location" so it is a clear instruction. 

I can see why you made this change. If you'd prefer, I think it's fine for this header to change when switched to manual data mode.

Whatcha think?

blanchardjeremy:
> Are you sure? I can't see this behaviour on microcovid.org. But it is simple to add. The change would be from:

Oh you're right. It's not that way on the live site.... Huh. I swore I got that behavior last night! *shrug*  Yeah, it's not a blocking issue, but my two cents is that we should include it. 

So yeah. I think we're ready to go after my comment about the "Step 1" header above.

Thanks for all you've done! 🙌

blanchardjeremy:
@david-mears-dfe I'd love to merge this into main before merging #646 so that you don't have extra work to do. If you have a chance, could you make the change to "step 1" and sync it with the base branch? Then I can merge it into main!

beshaya:
Landed as #656

beshaya:
Thanks for the PR, @david-mears-dfe! I'm really sorry that took so long.

david-mears-dfe:
Nice, thanks both.

blanchardjeremy:
Yeah. Thanks for both writing the code and for the feedback. 👍✅

PR #334, opened 2020-10-11
Calculator usability upgrade

beshaya:


src/components/calculator/SelectControl.tsx
```diff
@@ -25,12 +25,36 @@ function showRiskMultiplier(multiplier: number): string {
   }
 }
 
+interface LabelFormat {
+  label: string
+  header?: string
+  subLabel?: string
+}
+
+export const labelMaker = function ({
```
beshaya, line 34: Maybe just make this a React Component? (see e.g. signature for GenericSelectControl).

Also, maybe call it something more descriptive, like LabelWithHeader?

Alternatively, this behavior could put put into GenericSelectControl.

src/pages/Calculator.tsx
```diff
@@ -139,54 +142,73 @@ export const Calculator = (): React.ReactElement => {
     </span>
   )
 
+  function FirstTimeUserIntroduction() {
```
beshaya, line 145: This could be its own component + file.

src/styles/App.scss
```diff
@@ -162,6 +156,16 @@ a {
       margin: 10px 0;
       background-color: $warning-fill;
     }
+    .calculator-params {
+      label {
+        @include media-breakpoint-up(lg) {
+          min-height: 3em;
```
beshaya, line 162: This doesn't work at all browser sizes:
image

Ok for now, I can see if I can figure out something with tables to make this more robust.

src/pages/Calculator.tsx
blanchardjeremy: Done. ✅

src/styles/App.scss
blanchardjeremy: Hm. Why tables? I thought the general thinking is table shouldn't be used for layout. That's what flexbox is for these days, I thought. Curious your thinking there.

src/components/calculator/SelectControl.tsx
beshaya: (done)

src/styles/App.scss
beshaya: I meant the <Row> and <Col> tags that come from bootstrap, not a literal table. But from the slack it sounds like the direction is a single column anyway, so this is moot.

blanchardjeremy: Yeah lets head toward 1 column. I'll do that in a separate PR though.

PR #335, opened 2020-10-14
Allow the calculator to display numbers closer to 1e6uCoV

beshaya:
* Compute risk as 1 - (1 - risk_each) ^ N if we would otherwise show 10%
* Note that this allows us to show "1 housemate with covid = 30%"
* Augment display functions to show more sigfigs when close to 1e6

Fixes: 331

### DIFF
--- a/src/components/calculator/PointsDisplay.tsx
+++ b/src/components/calculator/PointsDisplay.tsx
@@ -4,7 +4,7 @@ import { Link } from 'react-router-dom'
 
 import { GenericSelectControl } from './SelectControl'
 import Card from 'components/Card'
-import { ERROR_FACTOR, MAX_POINTS } from 'data/calculate'
+import { MAX_POINTS } from 'data/calculate'
 import {
   fixedPointPrecision,
   fixedPointPrecisionPercent,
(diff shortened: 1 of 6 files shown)


blanchardjeremy (commented):
Slight bug (that maybe doesn't matter).
https://deploy-preview-335--microcov.netlify.app/?interaction=repeated&personCount=20&riskProfile=hasCovid&subLocation=US_06001&topLocation=US_06

result is 999,000 microCOVIDs

https://deploy-preview-335--microcov.netlify.app/?interaction=repeated&personCount=10&riskProfile=hasCovid&subLocation=US_06001&topLocation=US_06

result is 1,000,000 microCOVIDs

Not our main use case. But just a little odd.

beshaya:
image
image
image

mattbell:
Quick note -- it looks like the error range goes over 1m at the top.

beshaya:
> Quick note -- it looks like the error range goes over 1m at the top.
@mattbell Good point. I adjusted the error bound calculation code to apply the x3 or /3 to the per-person risk and then compute the repeated risk. It feels unnecessary, but quite pleasing to have numbers that make sense for very dangerous activities.

blanchardjeremy (approved):
Looks good to me! Thanks for adding/updating tests!

PR #352, opened 2020-10-22
Refactor Markdown page generation to be separate from paper.

beshaya:
This will make it easy to re-use this code to create the blog.

(There should be exactly no user-facing changes from this PR)


catherio (approved):
LGTM

PR #354, opened 2020-10-22
Adds a blog page!

beshaya:


blanchardjeremy:
You didn't request a review yet, but it looks good to me.

src/components/MarkdownPage.tsx
```diff
@@ -174,7 +179,7 @@ export const TableOfContents: React.FunctionComponent<{
       </div>
       {children}
       <ol className="toc">
-        {slugs.map((pageId, pageIndex) => (
+        {[...slugs].reverse().map((pageId, pageIndex) => (
```
blanchardjeremy, line 182: You might not be ready fora. review yet, so you might already know this, but: the reverse() here messed up the TOC for the white paper. It shows up in backwards order.

blanchardjeremy (changes requested):

src/components/MarkdownPage.tsx
beshaya: Good point. We can just control the ordering in the postMap.

PR #358, opened 2020-10-23
Refactor strong/bold to component & cleanup warning

blanchardjeremy:


src/components/calculator/PointsDisplay.tsx
```diff
@@ -33,6 +33,17 @@ function displayPercent(points: number): string {
   return showPoints(points) ? fixedPointPrecisionPercent(points * 1e-6) : '—%'
 }
 
+export function CalculationStepHeader(props: {
```
beshaya, line 36: Personally, I'd just make this a 2 variable function, since that allows single lines in the file below (just make it a function that returns a JSX object, call via `{calculationStepHeader()}`

blanchardjeremy: Done in #368

PR #359, opened 2020-10-23
Refactor ExplanationCard into a folder of files

beshaya:
PointsDisplay got wayyyy too big for one file. This breaks out all the pieces so that contributors can work on them independently.


blanchardjeremy (approved):
Looks good to me!

PR #360, opened 2020-10-23
Make advice for "close contacts" display for close contacts

beshaya:


beshaya:
(lgtm from Matt Bell)

Issue #341, opened 2020-10-17
Numbers in calculator no longer rounded


ddfisher:
The numbers in the calculator are no longer rounded to hide floating point errors.  This makes them somewhat confusing/harder to parse.
![image](https://user-images.githubusercontent.com/245905/96349708-38c3d980-1066-11eb-9c4b-3dc0e9af40f1.png)

blanchardjeremy:
@ddfisher can you link to the scenario you used to generate this?

blanchardjeremy:
Oh I found one. This scenario: https://www.microcovid.org/?distance=sixFt&duration=10&interaction=oneTime&personCount=5&riskProfile=livingAlone&setting=indoor&subLocation=US_06001&theirMask=filtered&topLocation=US_06&voice=normal&yourMask=filtered

![image](https://user-images.githubusercontent.com/32290/96350602-6f502300-106b-11eb-9518-b7fa72afb571.png)

dandv:
[Another one](https://www.microcovid.org/?distance=intimate&duration=600&interaction=oneTime&personCount=1&riskProfile=frontline&setting=indoor&subLocation=US_12011&theirMask=none&topLocation=US_12&voice=normal&yourMask=none)

> This is a roughly ~7,000-in-a-million (0.7000000000000001%) chance of getting COVID from this activity with these people.

blanchardjeremy:
Fixed with #395

PR #395, opened 2020-10-29
Fix fixed point formatting.

beshaya:
Closes: #341

### DIFF
--- a/src/components/calculator/util/FormatPrecision.ts
+++ b/src/components/calculator/util/FormatPrecision.ts
@@ -1,46 +1,80 @@
-const SIGFIGS = 1
-
-// Internal helper for rounding a number without inserting commas.
-function fixedPointNumber(val: number): string {
-  const orderOfMagnitude = Math.floor(Math.log10(val))
-  let sigfigs = SIGFIGS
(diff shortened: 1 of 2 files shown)


blanchardjeremy:
@beshaya  I don't know enough about this to give it a thorough code review but I can test the results. 

I guess my biggest question is: isn't there a library we could use for sug figs instead of rolling our own?

beshaya:
> @beshaya I don't know enough about this to give it a thorough code review but I can test the results.
> 
> I guess my biggest question is: isn't there a library we could use for sug figs instead of rolling our own? Just seems like a lot of code to have around.

I spent some time looking and couldn't find anything. The combination of wanting commas and no trailing zeros seems to fall in a dead spot for javascript. Adding in the fact that I want to display numbers close to 1e6 and percents close to 100 with arbitrary precision means we either write our own display code or do a lot of fiddling with someone else's library which wasn't built with that in mind :/ 

I might be able to get it a little shorter.

beshaya:
Took a stab at using more standard functions/libraries; the code got longer.

blanchardjeremy:
Thanks for doing this. And thanks for adding tests.

PR #396, opened 2020-10-29
Cleanup new user intro and refactor collapse into component.

### LINKED ISSUE #386: Improve the new user introduction
830BD67B-090E-435D-A6A1-B356988A6D09

blanchardjeremy:
Fixes #386 

(I know, the two refactor commits should probably be in a different PR. But I wanted to just knock them out on this branch. Let me know if you want me to separate them.)

### DIFF
--- a/src/components/Expandable.tsx
+++ b/src/components/Expandable.tsx
@@ -5,16 +5,26 @@ import { BsChevronDown, BsChevronRight } from 'react-icons/bs'
 export const Expandable: React.FunctionComponent<{
   id: string
   header: string
+  defaultState?: boolean
+  setter?: (value: boolean) => void
   headerExpanded?: string
+  headerClassName?: string
   className?: string
 }> = (props): React.ReactElement => {
-  const [open, setOpen] = useState(false)
+  const [open, setOpen] = useState(props.defaultState || false)
+
+  function setterHelper(isOpen: boolean) {
+    setOpen(isOpen)
+    if (typeof props.setter !== 'undefined') {
+      props.setter(isOpen)
+    }
+  }
 
   return (
-    <div className={`expandable-section ${props.className}`}>
+    <div className={`expandable-section ${props.className || ''}`}>
       <span
-        className="expandable-header"
-        onClick={() => setOpen(!open)}
+        className={`expandable-header ${props.headerClassName || ''}`}
+        onClick={() => setterHelper(!open)}
         aria-controls={props.id}
         aria-expanded={open}
       >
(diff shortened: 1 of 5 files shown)


src/components/Expandable.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { useState } from 'react'
+import { Collapse } from 'react-bootstrap'
+import { BsChevronDown, BsChevronRight } from 'react-icons/bs'
+
+export const Expandable: React.FunctionComponent<{
+  id: string
+  header: string
+  defaultState?: boolean
```
beshaya, line 8: can you use a default value here? I.e. defaultState = false

I'm not sure if it works in template types.

src/components/calculator/PrevalenceControls.tsx
```diff
@@ -231,124 +231,118 @@ export const PrevalenceControls: React.FunctionComponent<{
...
-        }
-        onClick={() => setDetailsOpen(!detailsOpen)}
-        aria-controls="prevelance-details"
-        aria-expanded={detailsOpen}
+      <Expandable
+        id="prevelance-details"
+        header="Details"
+        headerClassName={isManualEntryCurrently ? 'd-none' : ''}
+        defaultState={detailsOpen}
+        setter={setDetailsOpen}
```
beshaya, line 239: This doesn't work. If you click "manual entry" it no longer opens up the details, since setDetailsOpen has no way to propagate that to Expandable.

Instead, I think you want to break Expandable into two components:

ControlledExpandable has no internal state, and have `setter` and `isOpen` as mandatory variables.

Expandable wraps ControlledExpandable with  useState. It takes neither `setter` nor `isOpen`, but can take `defaultState`

In general, if there are two completely different ways a component can be managed, its probably a sign that it should be two different components.

beshaya (changes requested):
Probably change the PR title to reflect the refactor? This was a lot more code than I'd expect for "cleanup intro" :P

src/components/calculator/PrevalenceControls.tsx
blanchardjeremy: Thanks.  I knew there was a better way as I was writing it. I'll do it as you suggested!

src/components/Expandable.tsx
blanchardjeremy: I don't think so with template types.

src/components/calculator/PrevalenceControls.tsx
blanchardjeremy: Done!

beshaya (approved):
Thanks!

PR #403, opened 2020-10-29
Make points more prominent. 

### LINKED ISSUE #390: Make calculator result more prominent than the weekly budget
Per @catherio 

Right now the "200" is bigger than actual result. Which could be confusing. 

My preferred solution is to make the 200 smaller. And maybe make the calculator result a different color or something.

blanchardjeremy:
Fixes #390

### DIFF
--- a/src/components/calculator/PointsDisplay.tsx
+++ b/src/components/calculator/PointsDisplay.tsx
@@ -146,7 +146,7 @@ export default function PointsDisplay(props: {
         <div className="points">
           {doShowPoints ? (
             <>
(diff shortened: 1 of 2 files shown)


beshaya (approved):

beshaya:
I don't personally like adding the "Total Cost" since I consider it to be strongly implied, but assume someone else asked for it.

blanchardjeremy:
@beshaya nah. We're doing to remove it. We just agreed on it in slack. I'll have an updated PR this evening.

PR #408, opened 2020-10-30
Add guidance saved scenario

### LINKED ISSUE #387: Add extra guidance when using a saved scenario
> 3. Is it easy at this point to insert the brainstormed slight extra bit of guidance about “Describe the activity”?

> Eg. to make it say “You selected ‘Plane ride’ above. Does the following description sound like the activity you had in mind? If no, change it until it sounds right.”

blanchardjeremy:
Fixes #387

### DIFF
--- a/src/components/calculator/SavedDataSelector.tsx
+++ b/src/components/calculator/SavedDataSelector.tsx
@@ -38,6 +38,7 @@ export const SavedDataSelector: React.FunctionComponent<{
       props.setter({
         ...props.currentData,
         ...foundData,
+        scenarioName: value,
       })
     }
   }
--- a/src/data/calculate.ts
+++ b/src/data/calculate.ts
@@ -34,6 +34,9 @@ export interface CalculatorData {
   theirMask: string
   yourMask: string
   voice: string
+
+  // Scenario
+  scenarioName?: string
 }
 
 export const defaultValues: CalculatorData = {
@@ -54,6 +57,8 @@ export const defaultValues: CalculatorData = {
   theirMask: '',
   yourMask: '',
   voice: '',
+
+  scenarioName: '',
 }
 
 interface CalculatorResult {
--- a/src/pages/Calculator.tsx
+++ b/src/pages/Calculator.tsx
@@ -179,6 +179,16 @@ export const Calculator = (): React.ReactElement => {
                 <header id="activity-risk">
                   Step 2: Describe the activity
                 </header>
+                {!calculatorData.scenarioName ? null : (
+                  <div className="mb-3">
+                  <em>
+                    You selected this scenario:{' '}
+                    <strong>{calculatorData.scenarioName}</strong>. Do the
+                    points below look like the activity you have in mind? If
+                    not, adjust it until it looks right.
+                  </em>
+                  </div>
+                )}
                 <div>
                   <GenericSelectControl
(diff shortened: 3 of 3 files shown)


src/data/calculate.ts
```diff
@@ -54,6 +57,8 @@ export const defaultValues: CalculatorData = {
   theirMask: '',
   yourMask: '',
   voice: '',
+
+  scenarioName: '',
```
beshaya, line 61: unnecessary, since it will default to undefined.

beshaya: default to undefined?

src/components/calculator/SavedDataSelector.tsx
```diff
@@ -38,6 +38,7 @@ export const SavedDataSelector: React.FunctionComponent<{
       props.setter({
         ...props.currentData,
         ...foundData,
+        scenarioName: value,
```
beshaya, line 41: ResetForm doesn't currently clear the scenario dropdown, which seems wrong; could you fix that? I think the most straightforward way is use a GenericSelectControl below.

beshaya (changes requested):

src/data/calculate.ts
blanchardjeremy: We use this to reset the form when someone presses the reset button. So I think it is necessary for that case. Am I missing something?

src/components/calculator/SavedDataSelector.tsx
blanchardjeremy: Yep. Will do. I've got an open issue about that. Now seems like a good time to fix it.

blanchardjeremy: GenericSelectControl is meant to be used for selects that have a multiplier attached. 

I'm also not clear how using a generic select control would cause the form to reset when you press the reset button. How do those two connect?

beshaya: Mmm, yeah, GenericSelectControl wants the options to be FormValue.

I guess I don't much care whether you use GenericSelect or update this select (I thought GenericSelect would reduce the amount of boilerplate around styling, but maybe that's not true). The relevant part is setting `value` from `calculatorData` so that updating `calculatorData.scenario` updates the dropdown.

blanchardjeremy: Oh I see. You're saying it should be a two-way connection. Is that right?

Hmmm. I was treating it as a one-way connection where the calculator just holds the current label of the scenario. It's not attempting to hold the actual "scenario" (which would include storing the scenario key). If we want to do that, we could. Is that what you'd suggest?

What I was thinking is just adding some code to the reset button that clears the scenario dropdown. Would that work too?

beshaya: That doesn't make sense if scenarioName is part of calculatorData. Either it's part of calculatorData, in which case it should be controlled by calculatorData, or it's not, in which case it should NEVER be part of it.

I'm fine with either. I slightly prefer not putting it into calculatorData (since that will cause share-by-link to show "you selected X scenario", which the recipient did not.

blanchardjeremy: Ah got it. I'll remove it from calculator data then.

PR #411, opened 2020-10-30
Add risk budget to data model

### LINKED ISSUE #399: Risk budget not saved after page refresh
Occurs on the current v1 as well as the `ux-redesign` branch.

Replicate by: setting up an example scenario. Change your risk budget to "0.1%". Refresh Page.

Actual result: Risk budget goes back to "1%"

Expected result: Risk budget stays at "0.1%"

We need to move the risk budget into the URL params and the Calculator data model.

blanchardjeremy:
Fixes #399

### DIFF
--- a/src/components/calculator/ExplanationCard/ExplanationCard.tsx
+++ b/src/components/calculator/ExplanationCard/ExplanationCard.tsx
@@ -18,15 +18,13 @@ import { fixedPointPrecisionPercent } from 'components/calculator/util/FormatPre
 import Card from 'components/Card'
 import Expandable from 'components/Expandable'
 import {
-  BUDGET_CAUTIOUS,
-  BUDGET_STANDARD,
   CalculatorData,
   MAX_ACTIVITY_RISK,
   calculateActivityRisk,
   calculateLocationPersonAverage,
   calculatePersonRiskEach,
 } from 'data/calculate'
-import { intimateDurationFloor } from 'data/data'
+import { budgetOptions, intimateDurationFloor } from 'data/data'
 
 const calculationStepHeader = (header: string, value: string): JSX.Element => {
   return (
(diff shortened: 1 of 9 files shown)


src/styles/App.scss
```diff
@@ -360,6 +360,7 @@ input[type=number] {
 
     // Thermometer Credit. Adapted from: Copyright (c) 2020 COVID-19 Projections. https://github.com/covid-projections/covid-projections/
 
+    $thermometer-levels: 6;
```
beshaya, line 363: Looks like this file slipped in from your other change?

src/components/calculator/ExplanationCard/ExplanationCard.tsx
```diff
@@ -232,14 +233,14 @@ export default function ExplanationCard(props: {
       sublabel:
         'Budget: 1% chance of COVID per year (suggested for healthy people NOT in close contact with more vulnerable people)',
       multiplier: 1,
-      value: '10000',
+      value: BUDGET_STANDARD.toString(),
```
beshaya, line 236: Please move budgetOptions into data.ts, since it is an enum for calculatorData.

src/data/__tests__/calculate.test.ts
```diff
@@ -43,6 +47,7 @@ describe('calculate', () => {
     const scenario = 'Outdoor masked hangout with 2 people'
     const data: CalculatorData = {
       ...exampleLocation,
+      ...standardRiskBudget,
```
beshaya, line 50: it would be clearer to just put  `riskBudget: BUDGET_STANDARD` in each of these, rather than defining a single key object and then spreading it.

Even better, put riskBudget into exampleLocation (possibly rename exampleLocation to baseTestData) so you don't add overhead for new tests.

src/data/calculate.ts
```diff
@@ -36,7 +39,12 @@ export interface CalculatorData {
   voice: string
 }
 
+export const BUDGET_STANDARD = 10000
```
beshaya, line 42: These should be called BUDGET_ONE_PERCENT and BUDGET_TENTH_PERCENT so that it is obvious that changing the numeric value requires changes elsewhere (changing the text).

beshaya (changes requested):

src/styles/App.scss
blanchardjeremy: Woops. That's weird. I'll fix it.

src/data/__tests__/calculate.test.ts
blanchardjeremy: Oh good. Yeah I like the baseData solution. That was my first thought but I didn't want to mess with what was already there. Thanks. Will do.

src/data/calculate.ts
blanchardjeremy: Good call.

src/styles/App.scss
beshaya: Still here?

beshaya (approved):
Looks good, except the stylesheet changes are still there?

blanchardjeremy:
Closing in favor of #418

PR #434, opened 2020-10-31
Add empty optgroup to end of selects

beshaya:
This makes ios display multiline options.

### DIFF
--- a/src/components/calculator/SavedDataSelector.tsx
+++ b/src/components/calculator/SavedDataSelector.tsx
@@ -2,6 +2,7 @@ import { map } from 'lodash'
 import React from 'react'
 import { Form } from 'react-bootstrap'
 
+import IosOptgroup from 'components/IosOptgroup'
 import { CalculatorData } from 'data/calculate'
 import { PartialData, prepopulated } from 'data/prepopulated'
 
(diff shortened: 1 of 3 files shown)


src/components/calculator/SavedDataSelector.tsx
```diff
@@ -47,6 +47,7 @@ export const SavedDataSelector: React.FunctionComponent<{
           Optional: Start with a predefined common activity...
         </option>
         {prepopulatedOptions}
+        <optgroup label=""></optgroup>
```
blanchardjeremy, line 50: Could you add comments to explain the purpose of these? I found one the other day and removed it because I thought it was a bug.

blanchardjeremy (changes requested):
One change.

src/components/calculator/SavedDataSelector.tsx
beshaya: ahahahaha. Oopsies.

beshaya: Did one better - its now in a documented component that only renders on mobile ios.

blanchardjeremy (approved):

PR #435, opened 2020-10-31
Add copy to spreadsheet button. Fixes #432

### LINKED ISSUE #432: Add a "copy these values to the spreadsheet" button for prevalence data
Until #274 is done, we could have a stop-gap measure of offering a "copy these values to the spreadsheet" button that makes it easy to copy-paste an entire row into the Locations tab of the spreadsheet

blanchardjeremy:
Fixes #432


beshaya (approved):
LGTM.

PR #444, opened 2020-11-01
remove opengraph url

beshaya:
- Remove og:url
- Change the url of the blog post.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -48,6 +48,11 @@ export const App = (): React.ReactElement => {
             />
             <meta property="og:image:width" content="400" />
             <meta property="og:image:height" content="400" />
+            <meta
+              property="og:url"
+              // Keep the query params so that we can generate preview images for calculator results.
+              content={window.location.href}
+            />
             <meta property="fb:app_id" content="1117003835468995" />
             <meta property="og:type" content="website" />
             <title>microCOVID Project</title>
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -48,10 +48,6 @@ export const App = (): React.ReactElement => {
             />
             <meta property="og:image:width" content="400" />
             <meta property="og:image:height" content="400" />
-            <meta
-              property="og:url"
```
blanchardjeremy, line 52: Why not just add the full path to the page.  Looks like you're just adding the domain here instead of the full path to this page. Which explains the behavior you were getting. Wanna try re adding it with the full path?

blanchardjeremy (changes requested):
See above. 

If not, we can publish it as is

src/App.tsx
beshaya: If we don't include the query params, then facebook will break sharing scenarios by link. If we do include the query params as the og:url, then we're defeating the purpose of og:url (which is to enforce sharing clean urls)

beshaya: Hm, I guess just plopping in window.location.href would unify www vs no www and include the method, which is somewhat useful

beshaya: Done

blanchardjeremy (approved):

PR #448, opened 2020-11-01
Add custom entry params to share url.

beshaya:
Closes: 447


blanchardjeremy (approved):
Just did some basic testing trying to break it by putting conflicting sets of data in the URL and it still worked fine. Looks good to me.

PR #449, opened 2020-11-02
Add "200 microcovids per week" to radio buttons

catherio:
Now that we've removed "200 microCOVIDs per week" in the main calculator results, it'd be great if it was more visible and not just buried in "How was this calculated?"


blanchardjeremy (approved):
LGTM

beshaya (approved):

PR #459, opened 2020-11-02
Use paperPage class on About, Spreadsheet and Contact

dominic:
An ID was used in its place incorrectly.


blanchardjeremy (approved):
LGTM

PR #460, opened 2020-11-02
Scroll to top of screen when resetting

### LINKED ISSUE #456: Scroll to top of calculator after reset
It would be a little nicer for the user if we scroll tot he top of the calculator after a rest, otherwise they are left seeing this:

image

dominic:
Fixes #456


src/pages/Calculator.tsx
```diff
@@ -66,6 +66,7 @@ export const Calculator = (): React.ReactElement => {
   }
 
   const resetForm = () => {
+    window.scrollTo(0, 0)
```
blanchardjeremy, line 69: My two cents is that it would be better to scroll to the top of the calculator section rather than the top of the page. But if you wanna keep it this way, let's just merge it.

blanchardjeremy (approved):
Merge when ready.

PR #462, opened 2020-11-02
new categories

beshaya:
- Change closed pod labels to reflect hrs of grocery shopping.
- Use different multipliers for housemates' risk profile to avoid reflected risk.


beshaya:
image
image

Some notes - this assumes that partners live together. Obviously this is not always true. However, given that transmission risk has a narrow risk, there is good reason to believe that partners who don't see each other every day are at lower transmission risk anyway, so this probably balances out.

dominic (approved):
LGTM - needs a merge cause of i18n I bet.

PR #465, opened 2020-11-03
Fixing the note around mask usage and eating (#457)

### LINKED ISSUE #457: notes for both "Your mask" and "Their mask" refer to you eating or drinking
Why should my eating nullify their mask use? It nullifies my own mask use all right, but that's already taken care of by setting my mask usage to 'no mask'. If they're eating, it cancels out _their_ mask usage, but if they're wearing their mask, that protects me, even if I took mine off while I'm eating.

gyp:
This closes #457 and now merges cleanly onto the new main branch.

Both notes refer to your own mask, but others eating and drinking should only affect their own mask usage, not yours.


blanchardjeremy (approved):
LGTM

PR #471, opened 2020-11-04
I18n/hu small fixes

gyp:
Added a few improvements to the Hungarian translation based on feedback from others.


blanchardjeremy (approved):
I think we'll just approve any changes like this since you know best!

Issue #379, opened 2020-10-27
Add documentation in white paper for N95s


blanchardjeremy:
I'm pretty sure we don't have references in the Resource Sources section about why we chose 1/10 modifier for N95s.

#285 added the new category. I think any time we add an element to the calculator, we should include links to the literature that has us make that choice.

Taggin @beshaya

PR #475, opened 2020-11-05
Add a source for n95 masks. Fixes: #379

beshaya:


blanchardjeremy:
Just took a quick look (leaving the proper review to @catherio): Is there a reason you didn't capitalize "n95"?

src/posts/paper/14-research-sources.ts
```diff
@@ -54,6 +52,10 @@ We combine these numbers to estimate approximately a 4x reduction. Again, this i
 
 We note that we think masks on the infected person might not help proportionately as much at larger indoor distances, because surgical masks are most effective at stopping larger droplets ([Leung et al.](https://www.nature.com/articles/s41591-020-0843-2)).
 
+For n95 masks, we get our figure from [Jimenez et al.](https://docs.google.com/spreadsheets/d/16K1OQkLD4BjgBdO8ePj6ytf-RpPMlJ6aXFg3PrIQBbQ/edit#gid=519189277&range=C190:J190) who suggest that, while a perfect fit with an n95 is 99% efficient at filtering out aerosols, average use causes them to discount this to 90%.
+
+In the spreadsheet, we speculatively add professionally fit-tested n95 masks and P100 respirators (the latter because the soft silicone is more likely to have a good seal) at the full 99% filteration efficiency listed above.
```
oremanj, line 57: This is already discussed in the Q&A, not sure we should mention it here since it's not researched.

src/posts/paper/14-research-sources.ts
```diff
@@ -54,6 +52,10 @@ We combine these numbers to estimate approximately a 4x reduction. Again, this i
 
 We note that we think masks on the infected person might not help proportionately as much at larger indoor distances, because surgical masks are most effective at stopping larger droplets ([Leung et al.](https://www.nature.com/articles/s41591-020-0843-2)).
 
+For n95 masks, we get our figure from [Jimenez et al.](https://docs.google.com/spreadsheets/d/16K1OQkLD4BjgBdO8ePj6ytf-RpPMlJ6aXFg3PrIQBbQ/edit#gid=519189277&range=C190:J190) who suggest that, while a perfect fit with an n95 is 99% efficient at filtering out aerosols, average use causes them to discount this to 90%.
```
oremanj, line 55: N95 has a capital N.

Referencing Jimenez is a little awkward because we disagree with his other mask figures that are right next to that one. Chu might be a better source, or maybe there's something else we can dig up?

src/posts/paper/14-research-sources.ts
beshaya: Removed

src/posts/paper/14-research-sources.ts
beshaya: Chu says that N95 was about 10x better than other masks (0.04 vs 0.33), so we'd need to make the N95 multiplier 20x instead of 10x. Do you want me to do that? Or say we are discounting to 10x to account for fit?

beshaya: Done.

PR #479, opened 2020-11-07
Slight change to header for calculation details. Fixes #476

### LINKED ISSUE #476: Display the Person Risk for selected risk profile again?
Pre-UI-change, when I selected a risk profile (e.g. "In a closed pod of 4 people") from the dropdown, it would show the Person Risk below. I found this really useful, since it gave me the information I needed to compute slightly unusual scenarios (typically involving chains of people) that the calculator isn't built for. It looks to me like that line got removed in the makeover (specifically, in 9fb25b2838d251095d05b3d1b3b231bdc1c9befe, I believe).

Edit: upon inspection, it looks like this info was _deliberately_ removed, probably for decluttering reasons. But perhaps the information could get surfaced somewhere more out-of-the-way? I'll likely open a PR in the next couple weeks, unless I hear that you actively don't want this.

blanchardjeremy:
See #476


beshaya:
It looks fine code-wise, but does it actually improve discoverability? I don't think this closes that issue. Left a comment on the issue.

blanchardjeremy:
Okay I'll merge it and let's keep the issue open.

blanchardjeremy:
@gyp what approach should we use to notify translators that the source text has changed? Do you have any tools for that?

gyp:
@blanchardjeremy Truth be told, I never thought that far ahead. I only ever did this kind of localization in enterprise environments, so I don't really have anything battle-tested and ready-to-use for a small open-source project like this.

I see a number of options:

1. Configure a GitHub Action that automatically opens an issue for all currently supported languages on any change in src/locales/en.json, assuming that that is the original source, and assigning it to the maintainer responsible for that language. [This ready-to-use Action](https://github.com/JasonEtco/create-an-issue) seems like a good way to do that.
2. Integrate with a professional localization SaaS and make use of its features. [Crowdin](https://crowdin.com/pricing#crowdin-pricing-key-features) seems to be the one that has the best free option for a small project like this and they supposedly offer the enterprise features for OS projects if asked, but you might know and like others.
3. Self-host [Weblate](https://weblate.org/) or something similar and integrate with that. To be honest, it seems to be an overkill for a project of this size, but they have Docker-based images, so spinning it up and maintaining it should not be that problematic or expensive.

Let me know which one you prefer and I can start digging into it. My suggestion would be (1) for the time being, but (2) has the added benefit of making it much easier for non-tech contributors to participate.

PR #480, opened 2020-11-07
Update prevalence script to get it working

blanchardjeremy:
Cleans up a couple of bugs in the script.


update_prevalence.py
```diff
@@ -573,6 +576,8 @@ def parse_json(cache: DataCache, model: Type[Model], url: str) -> List[Model]:
 
 
 def ignore_jhu_place(line: JHUCommonFields) -> bool:
+    if line.Combined_Key == "":
+        return True
```
blanchardjeremy, line 580: This was one of the main things that got it working. Does this seem like a good addition to you?

blanchardjeremy:
@oremanj 

This is now an actual PR. I'm curious if these changes look good to you and whether we should merge them in so the script actually works.

update_prevalence.py
```diff
@@ -445,6 +447,7 @@ def rollup_testing(parent: Place, child_attr: str) -> None:
                             # These just don't have any reported cases
                             "Hoonah-Angoon, Alaska, US",
                             "Lake and Peninsula, Alaska, US",
+                            "Yakutat plus Hoonah-Angoon, Alaska, US",
```
oremanj, line 450: Huh, I wonder if this another "Bristol Bay plus Lake Peninsula" where the population is aggregated but the case data isn't. You can probably deal with it by adding an analogous special case around line 616.

update_prevalence.py
oremanj: I'd want to know which lines you're seeing that have an empty Combined_Key, but yes, seems like a fine workaround in principle.

update_prevalence.py
```diff
@@ -180,7 +180,8 @@ def recent_daily_cases(self) -> List[int]:
         daily_cases = []
         current = effective_date
         if current not in self.cumulative_cases:
-            raise ValueError(f"Missing data for {self.fullname} on {current:%Y-%m-%d}")
+            raise ValueError(
+                f"Missing data for {self.fullname} on {current:%Y-%m-%d}")
```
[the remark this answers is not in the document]
blanchardjeremy: Oh yeah. Thanks. I meant to fix this before committing. Let me see what the heck vscode is trying to do to it. I didn't intentionally set any formatted, so it's something that is preconfigured.

update_prevalence.py
blanchardjeremy: image

I get this when I run it without that line. It just appears to be a blank line as far as I can tell.

Let me know if you'd like any more info on this change before we accept it.

update_prevalence.py
blanchardjeremy: Okay. Fixed with a Bristol Bay-like solution.

update_prevalence.py
oremanj: Nah, that's enough for me!

Issue #492, opened 2020-11-09
Add HEPA filters as a ventilation option


beshaya:
Per conclusion of #314

beshaya:
BD4L says: 
![image](https://user-images.githubusercontent.com/5877421/98501033-f32ca380-2202-11eb-8456-69a6a50deeff.png)

PR #493, opened 2020-11-09
Add advanced ventilation options.

beshaya:
All 1/4:
* HEPA filters
* Planes/Trains
* Partially Enclosed

Closes: #492


src/locales/en.json
```diff
@@ -147,6 +147,10 @@
     "repeated": "Household member [{{ percentage }} chance of transmission per week]",
     "partner": "Partner / spouse [{{ percentage }} chance of transmission per week]",
     "indoor": "Indoor",
+    "filtered": "Indoors with a HEPA filter (flow rate 5x room size)",
+    "transit": "An airplane or train with air filtration",
+    "carWindowsDown": "A moving car with the windows rolled down",
```
blanchardjeremy, line 152: I noticed you used camel casing here. We're using underscores elsewhere.  It'd be good to keep it consistent. I don't care which one we use.

src/locales/en.json
```diff
@@ -147,6 +147,10 @@
     "repeated": "Household member [{{ percentage }} chance of transmission per week]",
     "partner": "Partner / spouse [{{ percentage }} chance of transmission per week]",
     "indoor": "Indoor",
+    "filtered": "Indoors with a HEPA filter (flow rate 5x room size)",
```
blanchardjeremy, line 150: I wonder how we can help people make sense of this. Maybe here by adding an info alert below the input box if they select this option. Or we could add it to the whitepaper and link them to it. 

In either case, I'd love to be able to give them some basic instructions. Even something like "Google the flow rate of your HEPA filter. Make sure it's rated XYZ or above. Multiply the with x length x height of your room. If the flow rate is greater than 5x that room size per hour, then you're good to go." (Except I'm butchering the explanation, but you get the idea.)

Whatcha think about adding something like that.

blanchardjeremy (changes requested):
Looks great overall. Thanks for doing this!! Couple of comments.

src/locales/en.json
beshaya: Yeah, I think a link to the whitepaper is better here.

src/locales/en.json
beshaya: Fixed

src/locales/en.json
beshaya: Done, ptal.

src/components/calculator/ActivityRiskControls.tsx
```diff
@@ -51,6 +51,15 @@ export const ActivityRiskControls: React.FunctionComponent<{
           <Trans>calculator.precautions.no_intimate_bonus_outdoors</Trans>
         </div>
       ) : null}
+      {data.setting === 'filtered' && (
+        <div className="warning">
+          <Trans>calculator.precautions.filtered_ensure_airflow</Trans>{' '}
+          <Link to="/paper/14-research-sources#ventilation">
+            <Trans>calculator.precautions.filtered_ensure_airflow_link</Trans>
```
blanchardjeremy, line 58: I've been learning way more about translation than I ever knew. One pro tip: don't assume grammer structure. Some languages might not end on the word "here". Instead, you can put the link inside a single <Trans> element. Then in the en.json file you can follow the pattern you see elsewhere of using <2>Foo</2> to fill in the link tag.

Let me know if you have any questions.

src/posts/paper/14-research-sources.ts
```diff
@@ -77,6 +77,21 @@ The sources above hint at up to a 100x benefit. Fellow armchair modeler Peter Hu
...
+* For a normal room height (8ft), ensure \`CFM > 2/3 * length * width\`
+
+The derivation of this rule is as follows:
+\`\`\`
+Refreshes/hr = 60 min/hr * CFM ft^3/min / (length ft * width ft * height ft) = 5
+60 * CFM / (length * width * height) = 5
+CFM = 5 * (length * width * height) / 60
+CFM = 1/12 * (length * width * height)
+if height = 8ft,
+CFM = 2/3 * (square footage)
```
blanchardjeremy, line 92: Thank you! Much more clear to users.

blanchardjeremy (changes requested):
One last change needed.

src/components/calculator/ActivityRiskControls.tsx
beshaya: Done! That's a really cool feature.

blanchardjeremy (approved):
Looks great.

PR #508, opened 2020-11-15
I18N: Weblate initial merge cleanup

gyp:
This merges in the initial changes from Weblate, most notably, adding fairly complete DE and FR translations.
They are not yet published, that will be done separately after everything is verified. It should not change anything visible on the site yet.

Weblate pulls started to fail due to the reorganization in en.json, so it'll probably have to be reset after this is merged in, but this is needed to make sure the translations are not lost.


blanchardjeremy (approved):
Thanks so much for getting this working!

PR #510, opened 2020-11-15
i18n: added a basic README for the translations

gyp:
The most important part is the first comment: we want to make sure people contribute via Weblate and not by making changes here.


blanchardjeremy (approved):
Brilliant!!

Issue #505, opened 2020-11-14
Displays "~.010 microcovids each time" instead of "~.10" sometimes


hmrm:
I'm guessing you've got some logic with the number of decimals to display and it doesn't account for rounding up, but I can't find it in the codebase.

blanchardjeremy:
@hmrm could you help us duplicate it by looking to a scenario where that shows up?

hmrm:
Currently https://www.microcovid.org/?distance=normal&duration=12&interaction=oneTime&personCount=1&riskProfile=nonFrontline&setting=outdoor&subLocation=US_53073&theirMask=basic&topLocation=US_53&voice=normal&yourMask=filtered displays the issue. Note that at 13 minutes the issue does not manifest

beshaya:
@hmrm Thanks so much for reporting this! Fix is in review.

PR #511, opened 2020-11-15
Fix display error when rounding a number < 1 increases the order of magnitude.

beshaya:
Closes: #505


src/components/calculator/util/__tests__/FormatPrecision.test.ts
```diff
@@ -22,6 +29,9 @@ describe('fixedPointPrecision', () => {
   it('shows precision for numbers close to 1e6', () => {
     expect(fixedPointPrecision(996700)).toEqual('997,000')
     expect(fixedPointPrecision(999920)).toEqual('999,920')
+    // This isn't super right, but it also seems odd to report 100% when it's not?
+    expect(fixedPointPrecision(1e6 - 0.1)).toEqual('999,999')
```
blanchardjeremy, line 33: If we round 99 to 100 why wouldn't we round this to 1,000,000?

blanchardjeremy (approved):
Looks good. One question.  You decide if you wanna address it.

src/components/calculator/util/__tests__/FormatPrecision.test.ts
beshaya: We round 99 to 100, but we _don't_ round 99% to 100%
expect(fixedPointPrecisionPercent(0.9992)).toEqual('99.92%')

blanchardjeremy: Oh I see

beshaya: Come to think of it, we should actually "round" 1e6 to 999,999 - because I never want to say something is certain.

PR #516, opened 2020-11-16
Help users find ventilation/filtration options

catherio:


catherio:
I believe “Environment: Is it indoor or outdoor?” used to be called “Ventilation”. When did this change? Is there a specific reason I should be aware of? People now cannot find the options for ventilation/filtration. https://twitter.com/Shawzie5/status/1327457924860547074

src/locales/en.json
```diff
@@ -11,7 +11,7 @@
     "precautions": {
       "label": "Precautions",
       "environment_header": "Environment",
-      "environment_question": "Is it indoor or outdoor?",
+      "environment_question": "Is it indoor or outdoor? Ventilation? Air filters?",
```
beshaya, line 14: Can we use full sentences to match elsewhere?

"Is it indoor or outdoor? Is there air flow or air filters?

beshaya (changes requested):

src/locales/en.json
catherio: cool, makes sense. slight change to keep "ventilation" visible/searchable on page

beshaya (approved):

PR #529, opened 2020-11-19
Emphasize protecting others

catherio:
Add "The first step in protecting your family and community is to avoid getting COVID yourself."


catherio:
Not waiting for a reviewer because got thumbs on slack from Jeremy for a description of exactly what I was going to do.

PR #540, opened 2020-11-24
Add discreet string types to CalculatorData interface

robinovitch61:
Hey all! Awesome site.

I cloned the project locally and was looking to run the `calculate` function from my terminal. I had to do a bit of searching around to see the options available for certain fields e.g. `riskProfile`. I copied the available options over from `data.ts` to make discreet string types for these fields.

May not be following DRY best-practices by doing this - very reasonable to not merge this small change in the name of maintainability. Just wanted to throw a small PR at a project I think is great! Best,

Leo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 .env.production.local
 .venv
 .prevalence_data.json
+.idea
 
 npm-debug.log*
 yarn-debug.log*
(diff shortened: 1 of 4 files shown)


beshaya:
Interesting idea. I think it would be cleaner to use keyof as the type for strings in CalculatorData.

blanchardjeremy:
Closing this as I think it's not the direction we're interested in taking things right now.

PR #564, opened 2020-12-09
Update Adjusted Prevalence Model

beshaya:
Use the model developed by https://covid19-projections.com/estimating-true-infections-revisited/

### DIFF
--- a/src/components/calculator/PrevalenceControls.tsx
+++ b/src/components/calculator/PrevalenceControls.tsx
@@ -374,6 +374,7 @@ interface PrevalanceData {
   casesPastWeek: number
   casesIncreasingPercentage: number
   positiveCasePercentage: number | null
+  prevalanceDataDate: Date
 }
 
 function dataForLocation(location: string): PrevalanceData {
@@ -389,6 +390,7 @@ function dataForLocation(location: string): PrevalanceData {
         locationData.positiveCasePercentage === null
           ? null
           : Math.round(locationData.positiveCasePercentage * 10) / 10,
+      prevalanceDataDate: new Date(PrevalenceDataDate),
     }
   }
 
@@ -397,5 +399,6 @@ function dataForLocation(location: string): PrevalanceData {
     casesPastWeek: 0,
     casesIncreasingPercentage: 0,
     positiveCasePercentage: 0,
+    prevalanceDataDate: new Date(),
   }
 }
--- a/src/components/calculator/PrevalenceControls.tsx
+++ b/src/components/calculator/PrevalenceControls.tsx
@@ -11,7 +11,6 @@ import {
   CalculatorData,
   calculateLocationPersonAverage,
   calculateLocationReportedPrevalence,
-  prevalanceRatio,
 } from 'data/calculate'
 import { TOP_LOCATION_MANUAL_ENTRY } from 'data/data'
 import { Locations, PrevalenceDataDate } from 'data/location'
@@ -280,15 +279,6 @@ export const PrevalenceControls: React.FunctionComponent<{
           isEditable={!locationSet}
           className="hide-number-buttons"
         />
-        {locationSet && (
-          <div>
-            Underreporting factor calcualted:{' '}
-            {prevalanceRatio(
-              data.positiveCasePercentage,
-              data.prevalanceDataDate,
-            ).toFixed(2)}
-          </div>
-        )}
         {locationSet && data.casesIncreasingPercentage === 0 ? (
           <div>{t('calculator.prevalence.cases_stable_or_decreasing')}</div>
         ) : (
--- a/src/components/calculator/PrevalenceControls.tsx
+++ b/src/components/calculator/PrevalenceControls.tsx
@@ -11,6 +11,7 @@ import {
   CalculatorData,
   calculateLocationPersonAverage,
   calculateLocationReportedPrevalence,
+  prevalanceRatio,
 } from 'data/calculate'
 import { TOP_LOCATION_MANUAL_ENTRY } from 'data/data'
 import { Locations, PrevalenceDataDate } from 'data/location'
@@ -279,6 +280,15 @@ export const PrevalenceControls: React.FunctionComponent<{
           isEditable={!locationSet}
           className="hide-number-buttons"
         />
+        {locationSet && (
+          <div>
+            Underreporting factor calcualted:{' '}
+            {prevalanceRatio(
+              data.positiveCasePercentage,
+              data.prevalanceDataDate,
+            ).toFixed(2)}
+          </div>
+        )}
         {locationSet && data.casesIncreasingPercentage === 0 ? (
           <div>{t('calculator.prevalence.cases_stable_or_decreasing')}</div>
         ) : (
--- a/src/data/__tests__/calculate.test.ts
+++ b/src/data/__tests__/calculate.test.ts
@@ -24,7 +24,7 @@ describe('calculate', () => {
     topLocation: 'mock state',
     label: 'mock city',
     population: '1,000,000',
-    casesPastWeek: 999, // will add 1 in pseudocount
+    casesPastWeek: 1999, // will add 1 in pseudocount
     casesIncreasingPercentage: 0,
     positiveCasePercentage: 1,
   }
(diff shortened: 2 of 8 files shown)


src/locales/hu.json
```diff
@@ -126,7 +126,7 @@
         "risk_tolerance_point1_percent_label": "Kockázatkerülő keret",
         "risk_tolerance_point1_percent_explanation": "Keret: évi 0.1% esély a fertőzésre, avagy 20 microCOVID hetente (javasolt érték, ha veszélyeztetett vagy, vagy gyakran találkozol veszélyeztetettel)",
         "intro": {
-            "whats_this": "<p>Ennek az oldalnak a segítségével meg tudod becsülni, hogy milyen eséllyel fertőződsz meg a koronavírussal különböző tevékenységek során. A <1>legjobb kutatási eredményeket</1> felhasználva hoztuk létre azzal a céllal, hogy segítsen használható intuíciót és megfelelő ártalomcsökkentési stratégiákat kialakítani, és csökkenteni a vírussal kapcsolatos stresszt.</p><p>A legfontosabb lépés a családod és a környezeted megóvására, ha te magad nem kapod el a koronavírust.</p>",
```
blanchardjeremy, line 129: @beshaya i think I remember @gyp saying that we shouldn't edit translation files directly. But I might be missing remembering. Can you double check with him?

catherio:
@beshaya I'd like to add some kind of (temporary) notice to the results box telling people

beshaya:
I think it should go with a blog post about high rates

src/locales/hu.json
gyp: Yes, exactly, as documented in the README file next to the translations:
https://github.com/microcovid/microcovid/blob/main/src/locales/README-TRANSLATIONS.md

We should not be editing translation files directly as it wreaks all kinds of havoc in the process. Weblate will, in theory, pick up the changes and remove no longer needed translations.

beshaya: Removed

catherio (approved):
LGTM

src/posts/paper/14-research-sources.ts
```diff
@@ -106,7 +106,7 @@ We get some corroboration for this number from the study of train passengers, [H
 
 ### Basic Method: Underreporting factor
 
-Previously we rolled out own estimation of underreporting factor based on raw caseloads. We now aim to have our cases numbers roughly follow those estimated by [covidestim](https://covidestim.org/) - a collaboration between Yale School of Public Health  and Harvard School of Public health. At the moment we have chosen underreporting factors that approximate their numbers, but are in the process of changing our injest process to use their numbers directly.
+Previously we rolled out own estimation of underreporting factor based on raw caseloads. We now aim to have our cases numbers roughly follow those estimated by [covidestim](https://covidestim.org/) - a collaboration between Yale School of Public Health  and Harvard School of Public health. At the moment we have chosen underreporting factors that approximate their numbers, but are in the process of changing our ingest process to use their numbers directly.
```
blanchardjeremy, line 109: There are two spaces between "Health" and "and":
`Public Health  and Harvard School`

beshaya: Outdated.

blanchardjeremy:
@beshaya what do you think about exposing the calculated underrporting factor here: 
image

blanchardjeremy:
Okay I went ahead and added it to the prevalence pane details. Let me know what you think.

src/components/calculator/PrevalenceControls.tsx
```diff
@@ -279,6 +280,15 @@ export const PrevalenceControls: React.FunctionComponent<{
           isEditable={!locationSet}
           className="hide-number-buttons"
         />
+        {locationSet && (
+          <div>
+            Underreporting factor calcualted:{' '}
```
beshaya, line 285: Typo.

Also, perhaps lets call this Prevalence Ratio, so its easy for someone reading COVID-19 Projections' page to understand what this is.

Also, if you do that, shouldn't you also expose the delay factor? Seems odd to expose one but not the other.

beshaya: I've reverted the commit. Having an underreporting factor is not a new feature, so I don't think displaying it should be a blocker for this PR. Please open a follow-on PR.

blanchardjeremy: Will do.

src/data/calculate.ts
```diff
@@ -69,6 +71,21 @@ interface CalculatorResult {
...
+
+export const DAY_0 = new Date(2020, 1, 12)
+const MS_PER_DAY = 1000 * 60 * 60 * 24
+
+// From https://covid19-projections.com/estimating-true-infections-revisited/
+const prevalanceRatio = (positivityPercent: number | null, date: Date) => {
+  const day_i = (date.getTime() - DAY_0.getTime()) / MS_PER_DAY
+  if (positivityPercent === null || positivityPercent > 100) {
+    positivityPercent = 100
+  }
```
blanchardjeremy, line 84: This is probably just a case of me not understanding, but it seems odd to me to have a null `positivityPercent` result in 100. It seems that the formula just results in "2" in that case. 

Maybe it'd be useful to add a comment to explain the intended result. Like "If we have no positivity data, we want to act as though... XYZ"

src/data/__tests__/calculate.test.ts
```diff
@@ -16,51 +21,108 @@ const calcValue = (data: CalculatorData) => {
...
-  const expectedPrevalance = 0.006
+describe('calculateLocationPersonAverage', () => {
+  it.each`
+    positiveCasePercentage | result
+    ${0}                   | ${RATE}
+    ${4}                   | ${BASE_RATE * (0.04 ** 0.5 * 25 + 2)}
+    ${6}                   | ${BASE_RATE * (0.06 ** 0.5 * 25 + 2)}
+    ${16}                  | ${BASE_RATE * (0.16 ** 0.5 * 25 + 2)}
+    ${100}                 | ${BASE_RATE * (25 + 2)}
+    ${null}                | ${BASE_RATE * (25 + 2)}
```
blanchardjeremy, line 70: Oh did you just add this from my slack comment or was this here before?

blanchardjeremy (changes requested):

src/data/__tests__/calculate.test.ts
beshaya: It was here before. Maybe you were thinking of casesIncreasingPercentage?

blanchardjeremy: I must have been looking at that set of tests. But I don't think it matters as much there, since there's no point in the logic outlined for what to do with a null case there, right?

src/data/__tests__/calculate.test.ts
```diff
@@ -1,12 +1,17 @@
...
   calculate,
   calculateLocationPersonAverage,
   defaultValues,
 } from 'data/calculate'
 import { BUDGET_ONE_PERCENT, RiskProfile, RiskProfileEnum } from 'data/data'
 import { prepopulated } from 'data/prepopulated'
 
+const MS_PER_DAY = 1000 * 60 * 60 * 24
+const BASE_RATE = 0.003
+const RATE = 0.006
```
blanchardjeremy, line 13: I don't understand the meaning of these or why they were chosen. Could you explain their purpose with a comment?

src/data/calculate.ts
blanchardjeremy: Thanks!

src/data/__tests__/calculate.test.ts
blanchardjeremy: So I'll call this resolved.

src/data/calculate.ts
beshaya: This is the same behavior as before - if there's no positive case percentage data, we assume the worst.

100% positivity rate -> `2 + 1250 / (day_i +25)`. Today, `day_i` is 300-something that's a little under 6x (so still less than the MINIMUM underreportingFactor previously.

In late summer (day 125 or so?) it would have been about 10x, which is what the underreportingFactor was for null positivityPercent (or positivityPercent over 15)

src/data/__tests__/calculate.test.ts
beshaya: Done

src/data/__tests__/calculate.test.ts
beshaya: In calculate.tsx, casesIncreasingPercentage is of type number, so it won't ever* be null.
positivityRate is `number?`, so it can be null.

Attempting to pass in `null` in the other test would cause the test to not compile. Putting `null` into the prevalence data would cause the build to fail.

Strong typing is good.

* you can totally make it null using unsafe casting (`as number`)

blanchardjeremy (approved):
LGTM. Still think @catherio should approve it.

PR #575, opened 2020-12-13
Break down RiskProfile calculation

beshaya:
- Create a feature branch for work to help with high prevalence situations + add to CI
- Breaks down RiskProfile calculation in individual's actions, # of housemate's, # of other traceable contacts; standardizes how RiskProfile is calculated
- Adds hooks for adding "symptom free" features

That this change makes no user-facing changes (except adding a new parameter in the url) - verified by calculate.test.ts

### DIFF
--- a/src/data/data.ts
+++ b/src/data/data.ts
@@ -7,10 +7,22 @@ export interface CheckBoxFormValue extends FormValue {
   sublabel?: string
 }
 
-export interface FormValue {
+export interface BaseFormValue {
   label: string
+  multiplier?: number
+}
+
+export interface FormValue extends BaseFormValue {
   multiplier: number
   housemateMultiplier?: number // Different multiplier to apply if the contact is a housemate.
+}
+
+export interface PersonRiskValue extends BaseFormValue {
+  label: string
+  personalMultiplier: number
+  numHousemates: number
+  numOtherTraceableContacts: number // Not including housemates
+  contactsMultiplier: number
 }
 
 const formValue = function (label: string, multiplier: number): FormValue {
@@ -112,6 +124,7 @@ export const budgetOptions = [
   },
 ]
 
+// TODO(beshaya): Move RiskProfile to it's own file.
 /*
  * Exposed to ten (silent distanced masked) average people indoors,
  * while wearing a surgical mask, one-time, for one hour per week.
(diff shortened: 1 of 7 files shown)


src/data/data.ts
```diff
@@ -112,11 +124,12 @@ export const budgetOptions = [
   },
 ]
 
+// TODO(beshaya): Move RiskProfile to it's own file.
```
blanchardjeremy, line 127: Do you wanna do that in this PR or are you saving it for a future one?

.github/workflows/ci.yml
```diff
@@ -4,9 +4,9 @@ name: CI
 # events but only for the master branch
 on:
   push:
-    branches: [ main, develop, ux-redesign ]
+    branches: [ main, feature/high-prevalence ]
```
blanchardjeremy, line 7: Do you wanna merge this PR into the new branch your proposing? Or into main (as it's set to right now)?

src/data/data.ts
```diff
@@ -125,69 +138,138 @@ const livingAloneMult =
   Interaction.oneTime.multiplier *
   Setting.indoor.multiplier
 
-export const RiskProfile: { [key: string]: FormValue } = {
+const SYMPTOM_FREE_LAST_SEEN_TODAY_MULT = 0.5
+const SYMPTOM_FREE_LAST_SEEN_MORE_THAN_THREE_DAYS_AGO = 1 / 7
+
+export const personRiskMultiplier: (arg: {
```
blanchardjeremy, line 144: So is this whole thing basically a refactor with some additional code architecture for future features? I wanna understand before I give it a thumbs up.

blanchardjeremy:
Also looks like there are some merge conflicts to resolve.

src/data/data.ts
beshaya: I'm intentionally not doing it in this PR because it would make the diff harder to read.

.github/workflows/ci.yml
beshaya: Mmph, I mean to land this in main because it doesn't change behaviors. This ci change can be excluded, though.

src/data/data.ts
beshaya: Correct. Before, person risks were opaque blobs of multiplication and addition. When I added `housemateMultiplier` it doubled down on this mess - two blobs of multiplication and addition that weren't obviously tied together. 

Right now, this pays down that debt - RiskProfile contains information about the # of housemates and traceable contacts for a risk profile, and all the multiplications and additions are handled by personRiskMultiplier. Therefore, if you trust  the logic of `personRiskMultiplier` (which, conveniently, is testable), then it becomes much easier to read and verify new RiskProfile categories, and it programmatically handles the case where the user is a housemate of the person in question.

The backend for the proposed additional feature this allows is included in personRiskMultiplier - you can specify that the contact and their contacts are verified symptom free to reduce their risk, which is a big player in making interactions safer (almost 4x for many categories).

Theoretically you could also use this to give credit for contact tracing (but the mechanics of asking for this are weird).

Happy to talk you through the change more if you'd like.

blanchardjeremy (approved):
LGTM. I didn't go over with a fine-toothed comb. But I did go over with a moderate-tooth comb. Haha. :)

PR #584, opened 2020-12-17
correction of translation CS

gromadan:
Translation of `restaurantIndoors` and `restaurantOutdoors` scenarios were switched


gromadan:
I've noticed that the translation is made using the weblate portal only afterwards. Now, I made the suggestion for correction there, so you can ignore this pull request I guess.

blanchardjeremy:
@gromadan thanks for helping fix it!

PR #585, opened 2020-12-18
Change distance comparison operator

### LINKED ISSUE #583: Change distances to be "6+ feet"
> I have gotten the feedback multiple times from folks that when they see ">6 feet" as an option (on the spreadsheet or the website) they think that if they are about 6 feet, they aren't allowed to use that category. They think they need to be like 7 feet or more. I wonder if we changed it to "6+ feet" instead if that would be clearer for users. What do you think? (I've got this on my checklist of things to consider before launching the new spreadsheet)

blanchardjeremy:
Change feet comparison operator. Clean up distance unit spacing and consistency

Fixes #583 

I debated many different ways of putting this:

* 6+ ft
* 6+ feet
* 6ft+

I'm open to doing it a different way. I think "6+ feet" is longer but clearer. I worry that "6ft+ / 2m+" will look too much like a math equation at first glance.

Looking forward to input!


IWalkInLA:
I agree that "6+ feet" is clearer. 

Re:: "6ft+ / 2+m"; are the units already converted/translated for countries that don't use freedom units? If not, I would lean towards this one; as it would make more sense to more of the world. However,  it does appear more "mathy"/intimidating (according to my  housemate whom has been struggling the most with the spreadsheet/microcovid process).

blanchardjeremy:
The units are only converted in regions where we have a complete translation available. We could probably have it automatically localize it to only read "6ft+" or "2m+" depending on what country the person is in. I'll open a separate issue for that.

beshaya (approved):

Issue #588, opened 2020-12-19
Erroneous Test Positivity Rate


beshaya:
All regions of CA on the Calculator show 3.3% tests coming back positive. This does not match covidactnow:

![image](https://user-images.githubusercontent.com/5877421/102682603-a633d980-417f-11eb-9de1-a74fa3177622.png)

![image](https://user-images.githubusercontent.com/5877421/102682608-b51a8c00-417f-11eb-8646-015b60f593b6.png)

This is getting imported into spreadsheets as well. The error is fairly significant; 15% test positivity rate raises the underreporting factor by 1.4 compared to 3%

IWalkInLA:
It's wrong for the entire US, per my post with #573 .

beshaya:
@IWalkInLA yup, I see that now. These are, incredibly, entirely separate bugs.

I'm fixing them both.

PR #589, opened 2020-12-19
Fix US Test Positivity Ratio imports

beshaya:
Fixes: #588


oremanj (approved):

update_prevalence.py
```diff
@@ -696,10 +697,12 @@ def main() -> None:
 
         # Test positivity per US county and state
         for item in parse_json(cache, CANRegionSummary, CANRegionSummary.COUNTY_SOURCE):
+            assert type(item.fips) is int, "Expected item.fips to be int but got {}".format(type(item.fips))
             if item.fips not in data.fips_to_county:
                 # Ignore e.g. Northern Mariana Islands
+                print(f"ignoring unknown county fips {item.fips}")
```
oremanj, line 703: I think there are several of these in normal operation -- not sure we should keep this printout?

beshaya: There are only 2 now that I've fixed the <5 digit FIPS code bug. Before the long list of these was (I think) every county in the first 10 states alphabetically (Alabama - Connecticut)

Issue #591, opened 2020-12-21
The character 'ő' looks strange on the site


gyp:
The character 'ő' ('o' with a double acute accent) is quite common in the Hungarian language, but it looks like there's no proper font for it in the font that's being used or there's an encoding problem -- see the attached screenshot.

![Screenshot 2020-12-21 112430](https://user-images.githubusercontent.com/1048356/102766943-2bd1a800-437f-11eb-9e4e-5940b401fd5d.png)

PR #595, opened 2020-12-21
style: include a more modern Lato font (Fixes: #591)

gyp:
A modern version supports more non-Latin characters that started to appear in translations.

License to use & redistribute the font is granted at
https://github.com/latofonts/lato-source/blob/master/LICENSE.txt


beshaya (approved):

beshaya:
Thanks for taking care of this!

PR #596, opened 2020-12-21
Update essential workers

beshaya:
* Increased risk is only for healthcare and social workers. 
* Risk reduced to 2x to reflect new research 
* Shows a thank-you for essential workers if this option is selected
* Removes "non essential" category 
* Shows risk in microCOVIDs in Risk Profile dropdown. 
* Tweak display strings for 1% and .1% risk budget groups


blanchardjeremy:
@beshaya i didn't see a review requested on this one yet, so I assume it's still a WIP?

src/locales/en.json
```diff
@@ -169,20 +170,20 @@
...
       "contact1": "Has 1 close contact (incl. roommates) who socializes but does not work. No other work or socializing.",
       "contact4": "Has 4 close contacts (incl. roommates) who socialize but do not work. No other work or socializing.",
       "contact10": "Has 10 close contacts (incl. roommates) who socialize but do not work. No other work or socializing.",
       "contactWorks": "Has 1 close contact (incl. roommates) who is an essential worker. No other work or socializing.",
-      "bars": "Goes to bars regularly",
-      "microcovid_budget_one_percent": "Uses microCOVID to maintain a risk of 1%/year (200 microcovid/week)",
-      "microcovid_budget_deci_percent": "Uses microCOVID to maintain a risk of 0.1%/year (20 microcovid/week)",
+      "bars": "Went to a bar in the last 10 days",
+      "microcovid_budget_one_percent": "Uses microCOVID to maintain a Standard Caution Budget",
+      "microcovid_budget_deci_percent": "Uses microCOVID to maintain a High Caution Budget",
```
blanchardjeremy, line 186: I wonder if this creates more confusion because people have to understand what those two budget categories mean. Which requires them to have experienced other parts of the app or documentation.

blanchardjeremy:
I let my housemate know about this new change (since I just made it for our house's spreadsheet) and she asked:

> What about people who work at grocery stores, or out in public in other ways?

This is a good question, I think. Especially for grocery store workers, which I imagine are above average. (I think folks like hair stylists and such could just be modeled as average). How would you suggest we address that? Two questions here I guess. 

1. What is the proper way to model them?
2. How do we communicate that?

beshaya:
Matt's research suggests these people are just average. I think this is
because, outside of weird bubbles like SF, the vast majority of people have
to go to work.

If you personally know a grocery worker, you can model them like anyone
else.

src/locales/en.json
beshaya: updated. lmk what you think.

blanchardjeremy: Looks good. Thanks.

src/locales/en.json
```diff
@@ -83,6 +83,7 @@
     "person_risk_profile_header": "Risk Profile",
     "household_members_note": "If you are modeling exposure from members of your household, only count their contacts outside the house to avoid double-counting",
     "large_group_warning": "Warning: This is a large number of people. Remember, you only need to include the number of people who are within 15 feet of you (not everyone present in the area). However, gathering a large number of people could put everyone at risk and creates the possibility of superspreader event.",
+    "frontline_worker_acknowledgement": "We deeply appreciate the sacrifices that essential workers are making during the pandemic. We encourage you to find safe ways of supporting these people, keeping them socially included and cared for. They should be significantly safer once vaccinated.",
```
blanchardjeremy, line 86: Should this be: "... the sacrifices that **frontline** workers are making.."

Seems to me that that term is more clear and matches with the "healthcare / social worker" label.

src/locales/en.json
```diff
@@ -170,20 +171,20 @@
     "loud_voice": "Loud talking (shouting, talking over music, singing)",
     "person": {
       "average": "An average person in your area",
-      "frontline": "An essential or front-line worker in your area",
-      "nonFrontline": "Someone in your area who is NOT an essential or front-line worker",
+      "frontline": "A healthcare or social worker",
+      "nonFrontline": "A random person in your area who works from home",
```
blanchardjeremy, line 175: Remind me: Why are we going with "random" instead of "average"?

I can see benefits to both. Just curious.

Also, seems like we should be consistent and use the same thing for both.

blanchardjeremy (approved):
Marking this as approved. I'm curious about your response to my last two comments, but neither of them are blocking.

src/locales/en.json
beshaya: shrugs. Changed to average.

I think "average" is a dumb word because literally no one is average.

src/components/calculator/ExplanationCard/ExplanationCard.tsx
```diff
@@ -88,6 +86,15 @@ export default function ExplanationCard(props: {
   const calculationBreakdown = () => {
     return (
       <>
+        <div id="changelog">
+          <Trans
+            i18nKey="calculator.explanationcard.last_updated"
+            values={{ date: lastUpdated.toLocaleDateString() }}
+          >
+            Model last updated on (date).{' '}
+            <Link to="/paper/changelog">(Change log)</Link>
```
blanchardjeremy, line 95: Changelog is one word I think?

src/posts/paper/8-intermediate-method.ts
```diff
@@ -1,40 +1,32 @@
...
-
+Finally, we note that Healthcare and Social Services workers are among the first who will be vaccinated. This should significantly reduce their risk of getting infected and is projected to be completed by the end of March.
 
+[^11]:
+     We derive this from [data from Washington State](https://www.doh.wa.gov/Portals/1/Documents/1600/coronavirus/IndustrySectorReport.pdf) that shows that Health Care and Social Assistance workers accounted for 25% of infections but just 13% of the population, or about twice as many cases as average. Most other categories, including Retail, Accomodation + Food Services, Construction, and Manufacturing were infected at the state average rate. Workers who could work from home, including Educational Services (Washington moved to distance learning) and Professional, Scientific, and Technical
```
blanchardjeremy, line 28: Thanks for adding this!

src/posts/paper/8-intermediate-method.ts
```diff
@@ -1,40 +1,32 @@
...
-
-[^essential]: Here are two definitions we have seen of “frontline” or “essential” workers. First: “Frontline workers include, but are not limited to, healthcare workers, protective service workers (police and EMTs), cashiers in grocery and general merchandise stores, production and food processing workers, janitors and maintenance workers, agricultural workers, and truck drivers” ([econofact.org](https://econofact.org/essential-and-frontline-workers-in-the-covid-19-crisis)). Second: “Essential workers are those who must leave their home to do their jobs AND: who interact in person with members of the public; OR who cannot maintain social distancing at their jobs; OR who work directly with people who are homeless or who have serious medical conditions or who are over age 60” (originally from [color.com](https://www.reddit.com/r/sanfrancisco/comments/gacw9v/covid19_testing_sites_falling_short_of_5000_test/)).
-
+Finally, we note that Healthcare and Social Services workers are among the first who will be vaccinated. This should significantly reduce their risk of getting infected and is projected to be completed by the end of March.
```
blanchardjeremy, line 25: Great note!

blanchardjeremy:
I say it's ready to merge when you're ready!

src/locales/en.json
blanchardjeremy: I updated this because I think it's more clear. Feel free to change back if you feel strongly about it.

mattbell:
Looks good, though I noticed one small typo in the whitepaper.  "If you also know that they work in healtcare or social services, then 5,000 * 2 = 10,000 so their Person Risk increases to 15,000." should be "If you also know that they work in healthcare or social services, then 5,000 * 2 = 10,000 so their Person Risk increases to 10,000."

blanchardjeremy:
@mattbell good catch. i just fixed that.

mattbell:
One more -- minor typo in changelog: "healtcare/social settings"

blanchardjeremy:
@mattbell fixed!

PR #599, opened 2020-12-26
Small cleanups

blanchardjeremy:


src/locales/en.json
```diff
@@ -222,4 +223,4 @@
     "copy_link": "Copy link to this scenario",
     "reset_form": "Reset form"
   }
-}
+}
```
beshaya, line 226: Why does your editor keep removing newlines? Please put it back 😛

beshaya (approved):

src/locales/en.json
blanchardjeremy: Haha. Good question. I'll find a setting for that. I guess our linter doesn't run on JSON files???

blanchardjeremy: K I found the setting. Won't happen again. Good catch.

PR #600, opened 2020-12-26
Remove copy to spreadsheet button

blanchardjeremy:
No longer needed. It was only a stopgap measure before we had automatic prevalence data imports.

### DIFF
--- a/src/components/calculator/CopyToSpreadsheetButton.tsx
+++ b/src/components/calculator/CopyToSpreadsheetButton.tsx
@@ -35,7 +35,9 @@ export const CopyToSpreadsheetButton: React.FunctionComponent<{
 
   const copySpreadsheetLocationData = () => {
     copy(generateLocationSpreadsheetRow(data), { format: 'text/plain' })
-    addAlert('Copied! Paste it into a row of the the "Locations" sheet.')
+    addAlert(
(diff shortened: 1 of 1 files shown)


beshaya:
Do we have a good way to communicate that folks using old sheets need to
upgrade to one with auto-importing?

blanchardjeremy:
@beshaya good question. It was only ever designed to work with a manual data entry for a very certain version of the spreadsheet. I don't think it works with old versions before that. So... *shrug* I'm assuming almost nobody uses it. But I don't have the data to back that up. And there's a chance I could be mis remembering.

Would you suggest we add a little notice for spreadsheet users for a while to encourage them to switch to the latest version?

beshaya:
> @beshaya good question. It was only ever designed to work with a manual data entry for a very certain version of the spreadsheet. I don't think it works with old versions before that. So... _shrug_ I'm assuming almost nobody uses it. But I don't have the data to back that up. And there's a chance I could be mis remembering.
> 
> Would you suggest we add a little notice for spreadsheet users for a while to encourage them to switch to the latest version?

Yeah, could we keep this button for, say, 2 weeks, and have it pop up a warning to upgrade?

beshaya:
ALSO WE SHOULD ADD ANALYTICS TO THAT BUTTON

blanchardjeremy:
Oops. Didn't mean to close this.

blanchardjeremy:
@beshaya ready for your review.

beshaya (approved):

PR #601, opened 2020-12-26
White Paper: No symptoms at time of interaction

blanchardjeremy:
Explicitly explain why we offer a 50% risk reduction if the person has no symptoms at the time you see them.


src/posts/paper/14-research-sources.ts
```diff
@@ -163,7 +163,9 @@ Another important thing to know about the infectious period is that only about 9
 
 ![He et al](${heExtended})
 
-(If you were previously familiar with this source, note that it was substantially updated in a [07 Aug 2020 author correction](https://www.nature.com/articles/s41591-020-1016-z).)
+*(If you were previously familiar with this source, note that it was substantially updated in a [07 Aug 2020 author correction](https://www.nature.com/articles/s41591-020-1016-z).)*
```
beshaya, line 166: Did you intend to italicize this?

src/posts/paper/14-research-sources.ts
```diff
@@ -163,7 +163,9 @@ Another important thing to know about the infectious period is that only about 9
...
+
+One insight we can gain is that since 50% of transmission occurrs after symptoms onset, if you confirm with the person you are seeing that they don't have symptoms right before you interact, you are reducing your risk by 50%.
```
beshaya, line 168: Is research sources a place people go for these things? Is this linked from anywhere? Otherwise I'd imagine FAQ would make more sense. Or possibly there should be a page for spreadsheet feature sources.

src/posts/paper/14-research-sources.ts
blanchardjeremy: Yeah. I like sidenotes like this to be in italics. But now that I think about it, it's actually better as a foot note.

src/posts/paper/14-research-sources.ts
blanchardjeremy: Yeah it's linked from a bunch of places, I think.

I could see this being in the FAQ as well. It just seems to fit here since all the surrounding context is here.

What would the FAQ question be?

src/posts/paper/14-research-sources.ts
blanchardjeremy: Okay I moved it to a footnote.

src/posts/paper/14-research-sources.ts
blanchardjeremy: Let me know if you'd like a change here.

beshaya (approved):

PR #603, opened 2020-12-27
Add mailchimp subscription form

blanchardjeremy:
So folks can opt in for updates about the spreadsheet.


beshaya (dismissed):
Could we just use a library? https://www.npmjs.com/package/react-mailchimp-form

blanchardjeremy:
Ah yeah I'm glad you didn't merge this because I want to use the existing Mailchimp account we have.

src/pages/Spreadsheet.tsx
```diff
@@ -63,7 +86,11 @@ export const Spreadsheet = (): React.ReactElement => {
         same thing as the {calculator}.
       </p>
 
-      <p>{spreadsheet('Use the spreadsheet here')}</p>
+      <p>
+        <strong>{spreadsheet('Use the spreadsheet here →')}</strong>
```
beshaya, line 90: I personally find the arrow weird; in my mind it should be pointing to something, like 

use the spreadsheet here -> <link>

I'd prefer it went away here and in the subscribe form, but won't block on it.

src/pages/Spreadsheet.tsx
```diff
@@ -16,6 +16,29 @@ function spreadsheet(text = 'spreadsheet') {
...
+        We will send email updates when we make feature upgrades and bug fixes
+        to the spreadsheet.
+      </p>
+      <p>
+        <a
+          href="http://eepurl.com/hb6y4T"
+          target="_blank"
+          rel="noreferrer"
+          className="btn btn-secondary"
+          role="button"
```
beshaya, line 33: I find it weird that this looks like a button but is just a link. Elsewhere our buttons are buttons and our links are links. Could you make this consistent?

beshaya (approved):
Approving with a style change request.

PR #611, opened 2020-12-30
Add middle seat empty option and fix full flight settings. 

### LINKED ISSUE #555: Flight Q&A doesn't match flight scenario
In the [Q&A](https://www.microcovid.org/paper/13-q-and-a#airplanes) we say 

> we’ve used 20 people at 6 feet of distance as a reasonable estimate for a moderately full flight.

But in the calculator when I select "Plane ride" it sets it to 20 people at 3 feet distance:
image

blanchardjeremy:
Fixes #555


beshaya (approved):
lgtm

PR #614, opened 2020-12-31
Tweak example scenarios 

### LINKED ISSUE #613: Grocery store scenario person count doesn't match
In the example scenario dropdown, a grocery store visit is 5 people at 6 feet away: https://www.microcovid.org/?distance=sixFt&duration=60&interaction=oneTime&personCount=5&riskProfile=average&setting=indoor&subLocation=US_06001&theirMask=basic&topLocation=US_06&voice=silent&yourMask=basic

In the rest of the system it's 10 people 6 feet away.

Seems like we should make these consistent, yeah? Unless there's a reason we're doing for 5 in that version?

blanchardjeremy:
* Add a second shopping scenario with few people. 
* Make grocery scenario have 10 people to match other parts of our system.  Fixes #613
* Change voting scenario to use a cotton mask rather than a filtered mask. We don't assume filtered masks anywhere else in the scenarios, so we shouldn't here, either.
* Re-order the example scenarios so the person-who-has-COVID ones are next to each other


beshaya (approved):
FWIW, all the RiskProfiles assume shopping is done with a filtered mask.

blanchardjeremy:
@beshaya yeah I think that should be changed as well because I don't think it's safe to assume that. 

I'll open a separate PR for that.

PR #621, opened 2021-01-06
Increase Activity Risk in response to B.1.1.7.

beshaya:
Apply 1.5x multiplier on one-time activities to account for B117 / VOC202019/01

### DIFF
--- a/src/data/__tests__/calculate.test.ts
+++ b/src/data/__tests__/calculate.test.ts
@@ -147,8 +147,8 @@ describe('calculate', () => {
     ${'Outdoor masked hangout with 2 people'}                       | ${9 * B117_CONTAGIOUSNESS_ADJUSTMENT}
     ${'Indoor unmasked hangout with 2 people'}                      | ${720 * B117_CONTAGIOUSNESS_ADJUSTMENT}
     ${'Car ride with 1 person for 15 mins'}                         | ${90 * B117_CONTAGIOUSNESS_ADJUSTMENT}
-    ${'One-night stand with a random person'}                       | ${2880 * B117_CONTAGIOUSNESS_ADJUSTMENT}
-    ${'Live-in partner who has no indoor interactions besides you'} | ${21.6 * B117_CONTAGIOUSNESS_ADJUSTMENT ** 2}
+    ${'One-night stand with a random person'}                       | ${2880}
+    ${'Live-in partner who has no indoor interactions besides you'} | ${21.6 * B117_CONTAGIOUSNESS_ADJUSTMENT}
     ${'Grocery store for 60 minutes (average # of shoppers)'}       | ${90 * B117_CONTAGIOUSNESS_ADJUSTMENT}
     ${'Grocery store for 60 minutes (few other shoppers)'}          | ${45 * B117_CONTAGIOUSNESS_ADJUSTMENT}
     ${'Plane ride (full flight)'}                                   | ${369 * B117_CONTAGIOUSNESS_ADJUSTMENT}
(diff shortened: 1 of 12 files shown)


beshaya:
Don't submit before filling in blog post body.

src/posts/blog/index.ts
```diff
@@ -2,10 +2,12 @@ import uxredesign from 'posts/blog/1-ux-redesign/ux-redesign'
 import thanksgiving from 'posts/blog/2-thanksgiving/thanksgiving'
 import hepafilters from 'posts/blog/3-hepa-filters/hepa-filters'
 import negativetests from 'posts/blog/4-negative-tests/negativeTests'
+import b117 from 'posts/blog/5-b117/b117'
 import { BlogPostMap } from 'posts/post'
 
 // Note: the keys in this map are the public URL for the post. Changing them will break links.
 export const pages: BlogPostMap = {
+  b117,
```
blanchardjeremy, line 10: This is kind of terrible SEO. Not that it matters much. But could we do something more friendly to the average non-technical user. Like `newcovidstrain` or something?

src/data/__tests__/calculate.test.ts
```diff
@@ -258,10 +260,11 @@ describe('calculate', () => {
       expect(calcValue(partner)).toEqual(calcValue(bonuses))
     })
 
-    it('should apply 48% risk', () => {
+    it('should apply 75% risk', () => {
```
blanchardjeremy, line 263: Would it be more accurate to put 72% here?

src/data/__tests__/calculate.test.ts
```diff
@@ -132,28 +134,33 @@ describe('calculate', () => {
...
-    ${'Large outdoor party: masked with 3 feet between people'}     | ${1080}
-    ${'Small indoor party: unmasked with 25 people'}                | ${27000}
-    ${'Outdoor, masked hangout with person who has COVID'}          | ${750}
-    ${'Indoor, unmasked hangout with person who has COVID'}         | ${60000}
-    ${'Voting in-person'}                                           | ${3}
+    ${'Outdoor masked hangout with 2 people'}                       | ${9 * B117_CONTAGIOUSNESS_ADJUSTMENT}
+    ${'Indoor unmasked hangout with 2 people'}                      | ${720 * B117_CONTAGIOUSNESS_ADJUSTMENT}
+    ${'Car ride with 1 person for 15 mins'}                         | ${90 * B117_CONTAGIOUSNESS_ADJUSTMENT}
+    ${'One-night stand with a random person'}                       | ${2880 * B117_CONTAGIOUSNESS_ADJUSTMENT}
+    ${'Live-in partner who has no indoor interactions besides you'} | ${21.6 * B117_CONTAGIOUSNESS_ADJUSTMENT ** 2}
```
blanchardjeremy, line 151: I don't totally understand why it's to the power of 2 instead of x2. But I believe you that it's accurate.

blanchardjeremy (approved):
Looks good to me. Minor changes requested, but nothing blocking.

blanchardjeremy:
@beshaya Please add a banner saying "we've adjusted things for the new COVID variant. read more here →"

src/data/__tests__/calculate.test.ts
beshaya: Live-in partner is calculated based on 1 hour of grocery shopping. Grocery shopping is 1.5x more dangerous, so Live-in partner is 1.5x more points. Then you have an interaction with live-in partner, which has 1.5x more activity risk because of b117.

old value = old person risk * old activity risk
new value = new person risk * new activity risk = (old person risk * 1.5) * (old activity risk * 1.5) = old value * 1.5 * 1.5

src/posts/blog/index.ts
beshaya: Is it? I expect folks to be searching B117 if they want to know about this one in particular. Also SEO should look at the page contents and title, which reference the new variant. Also "strain" is scientifically incorrect, SARS-CoV-2 is a strain, B117 is a variant. Also there are new variants all the time, this isn't the first or the last.

src/data/__tests__/calculate.test.ts
beshaya: updated

src/data/__tests__/calculate.test.ts
blanchardjeremy: Gotcha. could you add a comment explaining this since it's no totally obvious upon reading how the chain effects works.

src/posts/blog/index.ts
blanchardjeremy: Oh thanks for the strain vs variant FYI.

Okay I'm sold. Makes ense.

blanchardjeremy:
@beshaya Can you add the notice to the calculator page (using an `<Alert>`) that mentions the change and links to the blog post?

blanchardjeremy (changes requested):
A quick skim of this looks good to me. 

Only remaining change needed is the front page notice with a link to the blog.

src/data/__tests__/calculate.test.ts
beshaya: No longer applicable.

But... in general everything in the calculator is multiplicative. The only thing that is additive is # of people.

beshaya:
> A quick skim of this looks good to me.
> 
> Only remaining change needed is the front page notice with a link to the blog.

Done

Issue #620, opened 2021-01-05
When choosing combination of kissing and loud speaking the risk is multiplied


kubantjan:
When I at the same time select kissing and loud speaking those two scores get multiplied and I get extreme score.

I guess I should get this extreme score also if I only select kissing, because no matter how loud I speak the transmission is very likely

Or am I wrong? It certainly seems weird that kissing with hushed voices would mean so much less risk then kissing with shouting.

![Screenshot from 2021-01-05 22-08-08](https://user-images.githubusercontent.com/24394374/103699162-d2828f00-4fa2-11eb-8282-0a2a89176b46.png)

blanchardjeremy:
This is a good thing to notice. If it's kissing, it seems like we should just ignore the volume field altogether.

jy2wong:
Jeremy, can I counterpropose taking the worse of the volume and kissing multipliers? I think kissing + loud talking should be treated as loud talking and that kissing + no talking should be treated as kissing.

PR #644, opened 2021-01-14
Handle volume risk multiplier thoughtfully when kissing

jy2wong:
When kissing, take the worse of the kissing and volume multipliers. In
particular, this means that

* there is no bonus for not talking when you are kissing
* kissing and loud talking multipliers are not multiplied together when
  kissing


src/data/calculate.ts
```diff
@@ -251,21 +251,27 @@ export const calculateActivityRisk = (data: CalculatorData): number | null => {
     ): number => (given === '' ? 1 : table[given].multiplier)
 
     let effectiveDuration = data.duration
-    multiplier *= mulFor(Distance, data.distance)
     if (data.distance === 'intimate') {
```
beshaya, line 254: This would probably be easier to parse as a switch statement.

beshaya (changes requested):
Please add a test case or two in data/__tests__/calculate.test.ts

blanchardjeremy:
I guess I'm curious how there is such an activity as "kissing + loud talking" ... It seems like it should really be two activities: 1 that you log as kissing and the other that you log as loud talking. I'm curious what you have in mind when you say that you could select both. What's the example scenario?

Thanks!

jy2wong:
> I guess I'm curious how there is such an activity as "kissing + loud talking" ... It seems like it should really be two activities: 1 that you log as kissing and the other that you log as loud talking. I'm curious what you have in mind when you say that you could select both. What's the example scenario?
> 
> Thanks!

I didn't have one in mind to start with (just wanted to address #620), but if pressed, I'd suggest a karaoke night date.

beshaya:
It's not so much that there's an activity where you *should* select both as
the calculator *lets* you select both. Possibly we should just remove the
talking dropdown?

beshaya:
Catherio approved giving kissing the same multiplier as loud talking, if you want to include that here (ok in a follow-up as well)

blanchardjeremy:
My two cents: I'm in favor of removing the conversation volume when selecting kissing. And I'll let y'all make a final call.

jy2wong:
I removed the conversational volume question in the kissing scenario, but then the only precaution left is the indoor/outdoor/ventilation one, which doesn't even do anything (it just tells you that it's ignoring the value).

What about hiding the *entire* Precautions section if you're kissing?

blanchardjeremy:
That sounds right to me. I'd suggest we remove the warning about "outdoor doesn't apply when kissing" and re-add some notice that explicitly says "no precautions apply when kissing" ... so the user understands why everything else disappeared. 

Thoughts?

(Maybe just removing them is obvious enough)

jy2wong:
I don't think it's necessary to explain. I think new additions to the bare minimum reading path need to achieve/satisfy at least one of these:

* address surprise (make it easy to quickly answer natural questions)
* be a sufficiently common source of surprise to justify making everyone else read it
* drive helpful changes in behaviour
* build future intuition about how to model

I don't think kissing satisfies any of these.

beshaya:
Makes sense to me. The reason this was here in the first place was because
we didn't have kissing, only "cuddling" or "close" where the other
multipliers make sense.

jy2wong:
Current version complete removes precautions section including the header during the kissing scenario and reworks translation string to specifically be about the close case and not the intimate case. Also made it show up for the other ventilation-related bonuses.

PTAL?

src/locales/en.json
```diff
@@ -13,7 +13,7 @@
       "label": "Precautions",
       "environment_header": "Environment",
       "environment_question": "Is it indoor or outdoor? Is the space ventilated, or is the air filtered?",
-      "no_intimate_bonus_outdoors": "Due to very close distances, we are not confident that being outdoors reduces the risk in a substantial way. Thus, we are not providing any bonus for being outdoors when intimate.",
+      "no_close_bonus_environment": "Good environment choices are being ignored in the calculation. We are not confident that good ventilation or being outdoors strongly reduces the risk when people are this close to each other.",
```
beshaya, line 16: Kind of awkward wording?

"Good things are being ignored"

Also, changing the json key will cause all existing translations to show the English until they are re-translated. I think it is preferable to show the current translations (keep "no_intimate_bonus_outdoors")

beshaya (changes requested):

beshaya (approved):

Issue #645, opened 2021-01-14
Use a typeahead for long dropdowns


beshaya:
The length of the predefined activity list is currently limited by the fact that scrolling through all of them. A typeahead would make this searchable, so we could support many more activites.

beshaya:
Jenny found in a user study:
> TX has many, many counties. Type to search for county would be nice.

So probably use this for locations as well!

PR #646, opened 2021-01-14
Change locations and pre-defined activities to use a typeahead.

beshaya:
Closes: #645


beshaya:
image

image

blanchardjeremy:
Two thoughts: 

1) Why are the options underlined? Could we not underline them instead? I think it's unnecessary and visually distracting.

2) I have an idea. I'm not sold on the idea yet. But I wonder if it's wise for us to have labels above state/country & region?  I'm basing that on [this blog post](https://mono.company/design-practice/the-10-commandments-of-good-form-design-on-the-web/).

:)

beshaya:
> Two thoughts:
> 
> 1. Why are the options underlined? Could we not underline them instead? I think it's unnecessary and visually distracting.
> 2. I have an idea. I'm not sold on the idea yet. But I wonder if it's wise for us to have labels above state/country & region?  I'm basing that on [this blog post](https://mono.company/design-practice/the-10-commandments-of-good-form-design-on-the-web/).
> 
> :)

Both done.

src/components/calculator/PrevalenceControls.tsx
```diff
@@ -194,59 +199,79 @@ export const PrevalenceControls: React.FunctionComponent<{
...
+        return { label: Locations[locKey].label, value: locKey }
+      })
+  const selectedSubLocation = subLocationOptions.find(
+    (option) => option.value === data.subLocation,
+  )
+
   return (
     <React.Fragment>
       <header id="location">
         <Trans>calculator.location_selector_header</Trans>
```
blanchardjeremy, line 229: this broke

image

blanchardjeremy:
It's going to take a bit of work to merge this with #326. Any thoughts on which one we should do first?

beshaya:
> It's going to take a bit of work to merge this with #326. Any thoughts on which one we should do first?

The merge doesn't look bad to me; I can handle rebasing in either direction. Merge when ready.

src/components/calculator/PrevalenceControls.tsx
beshaya: Fixed! Thanks for the catch!

blanchardjeremy:
In that case, I suggest we merge #326 first (in the next few days) before merging this one.

blanchardjeremy:
Or, if you're willing to do the re-base for the other ticket, I'm down to merge this one first. I just don't wanna make extra work for @david-mears-dfe. He's already done so much on that ticket.

PR #657, opened 2021-01-21
Fix PrevalenceFields' display logic

david-mears-dfe:
PR was requested [here](https://github.com/microcovid/microcovid/pull/326#issuecomment-760600469).

When the location is set to 'Select location', and the Details drop-down is open, the fields were editable. The isEditable used the slightly wrong method which checks whether the top location has been set. This commit irons out that edge case.

### DIFF
--- a/src/components/calculator/PrevalenceControls.tsx
+++ b/src/components/calculator/PrevalenceControls.tsx
@@ -310,7 +310,7 @@ export const PrevalenceControls: React.FunctionComponent<{
             setter({ ...data, casesPastWeek: parseInt(value || '') })
           }
           inputType="number"
-          isEditable={!locationSet}
+          isEditable={isManualEntryCurrently}
           className="hide-number-buttons"
         />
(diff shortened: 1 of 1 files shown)


david-mears-dfe:
I've just thought that a better solution might be to hide the Details section entirely if the user is in location-choosing mode but has not chosen a location.

blanchardjeremy:
No strong preference on the currently coded solution vs the one you proposed in the comments. 

Just make sure you pull the latest code from `main`, since the new typeahead code is in place in the same areas you're editing.

david-mears-dfe:
I'll leave this as it is then.

david-mears-dfe:
Rebased on main.

blanchardjeremy:
I merged the other one. Since the branch is in your fork, I can't merge main in.

blanchardjeremy:
Closing in favor of #679 so I can easily get it up to date with main.

PR #658, opened 2021-01-21
Show results of prevalence calculation when in manual data mode

david-mears-dfe:
@blanchardjeremy [says](https://github.com/microcovid/microcovid/pull/326\#issuecomment-760626478):

> After the user types in the manual data, I think they should see the actual
prevalence and adjusted prevalence numbers still.

This commit removes the logic which hides the calculated prevalence numbers during manual data entry mode.


blanchardjeremy:
Perfect. Looks good.

Can you merge main into this branch?

david-mears-dfe:
Rebased on main.

PR #669, opened 2021-01-22
Move prebuilt activities to the person-risk card.

jy2wong:
It was previously below the call to action.

Make some related UX tweaks too:

* adjust header styles to deemphasize location and highlight
  activity/relationship now that location autofills
* remove explicit step numbers now that location autofills
* adjust prebuilt activities box instructions so that the most important
  words for understanding are clustered at the beginning of the phrase

* add magnifying glass affordance to prebuilt activity search to more
  strongly indicate that it's for typing (it's indistinguishable from
the other dropdowns. This isn't a perfect solution but will do for now.)

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -125,7 +125,7 @@
       "enter_data_manually": "Switch to manual data entry mode"
     },
     "prevalence_info_source_information": "Prevalence data consolidated from <1>Johns Hopkins CSSE</1> (reported cases), <4>Covid Act Now</4> (US positive test rates), and <7>Our World in Data</7> (international positive test rates).",
-    "select_scenario": "Optional: Start with a predefined common activity",
+    "select_scenario": "Search for an activity or build your own below...",
     "baseline_risk": "baseline risk",
     "risk_modifier_frac_2nd": "{{frac}} the risk",
     "risk_modifier_frac_3rd": "{{frac}}rd the risk",
--- a/src/components/calculator/PrevalenceControls.tsx
+++ b/src/components/calculator/PrevalenceControls.tsx
@@ -237,6 +237,7 @@ export const PrevalenceControls: React.FunctionComponent<{
         />
         <Typeahead
           clearButton={true}
+          highlightOnlyResult={true}
           id="top-location-typeahead"
           onChange={(e: Option[]) => {
             if (e.length !== 1) {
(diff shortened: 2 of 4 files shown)


src/locales/en.json
```diff
@@ -125,7 +125,7 @@
       "enter_data_manually": "Switch to manual data entry mode"
     },
     "prevalence_info_source_information": "Prevalence data consolidated from <1>Johns Hopkins CSSE</1> (reported cases), <4>Covid Act Now</4> (US positive test rates), and <7>Our World in Data</7> (international positive test rates).",
-    "select_scenario": "Optional: Start with a predefined common activity",
+    "select_scenario": "🔍 Search for an activity or build your own below...",
```
blanchardjeremy, line 128: This looks great to me overall. My only slight concern is with the use of an emoji here. 

**Option 1)** If we do go the emoji route, we should move it outside of the translation, so it doesn't rely on every single translator propegating that change.

**Option 2:** Personally, I'm leaning more in the direction of adding an icon. We can use any icon in [this library](https://react-icons.github.io/react-icons/icons). We've been tending toward the [bootstrap icons](https://react-icons.github.io/react-icons/icons?name=bs). There is one called `BsSearch`. We'd need to place it in the code, not in the translation file, of course. --- The only flaw it that since the text is a placeholder, I'm not 100% sure if the icon will render there properly (without a bunch of extra styling). It might only accept normal characters. 

Either option works for me. Your call.

blanchardjeremy (changes requested):

jy2wong:
I took the colour changes out of this particular PR and will incorporate them later. I think those are better grouped together elsewhere as general focus management improvements.

src/locales/en.json
jy2wong: Went with option 2. I'm not 100% thrilled with it, but I think "clicked on magnifying glass; slightly confused" is still an improvement on "I didn't know I could type here".

Making the magnifying glass internal to the typeahead box looks like it would involve an even more intense source code dive of react-bootstrap-typeahead.

blanchardjeremy:
Ah, I forgot about input group prepends in Bootstrap. Good use! Thanks for putting in that fix! I'll merge now.

blanchardjeremy (approved):

Issue #177, opened 2020-08-31
Request user location via Geolocation API


akrolsmir:
Step 1 of the calculator asks users to pick through a long dropdown (and US users to pick a second county as well) -- a faster UX flow would be to request the user's location directly e.g. 
https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API. Caveats:

1) Users might be reluctant to grant this permission, so you probably still want the backup option of choosing a default
2) Taking latlong the API returns, and mapping it to a specific country/state/county, would require some work (probably integrating another mapping API like https://developers.google.com/maps/documentation/geocoding/overview)

pde:
You can get a decent approximation using a [server-side lookup of the visitor's source IP address](https://www.npmjs.com/package/geoip-lite). Just be sure to note in the UI where it came from (and ideally for privacy, don't keep logs of those IPs for long if at all).

akrolsmir:
Oh, that's really cool! I don't know if there's actually a server component to this project; my impression (from 5 seconds of glancing) is that this is a pure frontend app hosted on Netlify, but I could be wrong (or there might be some way to decode the user's IP address client-side)?

oremanj:
Thanks for the suggestion! We discussed this and didn't wind up having the resources to implement it before release. There is currently no server-side component. The IP-based lookup is stateless, though, so it would be easy to add on using a different cloud service.

blanchardjeremy:
@benjaffe - I don't think we have a Google API key.

We'd need to consider how this change would intersect with https://github.com/microcovid/microcovid/issues/451.

blanchardjeremy:
Also, this isn't the same as #311, but it's along the same lines of making the location box easier, so I thought I'd cross-reference them.

benjaffe:
Seems like there are two options:
1. server-side lookup -- easy from the user's perspective, but could also be seen as creepy. "how did they know?"
2. client geolocation api -- requires two button presses... one for us to try and get their location, and two for them to accept location permissions from the browser itself. This might make it less of a concern with respect to #451 (though should still be mentioned in the policy)

PR #675, opened 2021-01-24
i18n: localized country selector

gyp:
This mostly works, but there are a few things to iron out before it could be merged:

- The "United States (all)" selection does not have the "(all)" postfix any more which can be confusing
- this would need to be merged immediately after a prevalence update patch to ensure the diff for `location.ts` is readable: as it's a generated file, manually staging just the relevant changes would infeasible.

Let me know your thoughts.

### DIFF
--- a/src/components/calculator/PrevalenceControls.tsx
+++ b/src/components/calculator/PrevalenceControls.tsx
@@ -212,8 +212,8 @@ export const PrevalenceControls: React.FunctionComponent<{
     false || isManualEntryCurrently,
   )
 
-  const topLocationOptions = Object.keys(locationGroups)
-    .flatMap((groupName) => {
+  const topLocationOptions = Object.keys(locationGroups).flatMap(
+    (groupName) => {
       return locationGroups[groupName].map((locKey) => {
         const country_label =
           Locations[locKey].iso3 &&
@@ -226,8 +226,14 @@ export const PrevalenceControls: React.FunctionComponent<{
             : Locations[locKey].label
         return { label: country_label, value: locKey }
       })
-    })
-    .sort((a, b) => a.label.localeCompare(b.label))
+    },
+  )
+
+  // reorder the list according to the current locale, but keep
+  // English as-is to make sure US states remain at the top
+  if (i18n.language !== 'en') {
+    topLocationOptions.sort((a, b) => a.label.localeCompare(b.label))
+  }
 
   const selectedTopLocation = topLocationOptions.find(
     (option) => option.value === data.topLocation,
--- a/src/components/calculator/PrevalenceControls.tsx
+++ b/src/components/calculator/PrevalenceControls.tsx
@@ -1,3 +1,5 @@
+import i18n from 'i18n'
+import countries from 'i18n-iso-countries'
 import { isNumber } from 'lodash'
 import React, { useEffect, useState } from 'react'
 import { Card, Form, InputGroup } from 'react-bootstrap'
@@ -139,6 +141,11 @@ export const PrevalenceControls: React.FunctionComponent<{
   setter: (newData: CalculatorData) => void
 }> = ({ data, setter }): React.ReactElement => {
   const { t } = useTranslation()
+  for (const iso_code of Object.keys(i18n.services.resourceStore.data)) {
+    countries.registerLocale(
+      require('i18n-iso-countries/langs/' + iso_code + '.json'), // eslint-disable-line @typescript-eslint/no-var-requires
+    )
+  }
   const locationGroups: { [key: string]: Array<string> } = {}
   for (const key in Locations) {
     const location = Locations[key]
@@ -208,7 +215,16 @@ export const PrevalenceControls: React.FunctionComponent<{
   const topLocationOptions = Object.keys(locationGroups).flatMap(
     (groupName) => {
       return locationGroups[groupName].map((locKey) => {
-        return { label: Locations[locKey].label, value: locKey }
+        const country_label =
+          Locations[locKey].iso3 &&
+          countries.getName(Locations[locKey].iso3!, i18n.language, {
+            select: 'official',
+          })
+            ? countries.getName(Locations[locKey].iso3!, i18n.language, {
+                select: 'official',
+              })
+            : Locations[locKey].label
+        return { label: country_label, value: locKey }
       })
     },
   )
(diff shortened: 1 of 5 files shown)


blanchardjeremy:
Thanks so much for doing this!

Would you like a review now or after it's ready to go (or both)?

gyp:
@blanchardjeremy A review now would be great, the "(all)" postfix should not add much new code -- and if it did, I'll ask for another round of reviews. I just want to know if I messed up something and should plan for further work on this or this is roughly OK in your view.

If we agree that the patch is OK, the two of us should coordinate on the prevalence update patch, I saw that it's you who tends to do those.

src/components/calculator/PrevalenceControls.tsx
```diff
@@ -1,9 +1,12 @@
+import countries from 'i18n-iso-countries'
 import { isNumber } from 'lodash'
 import React, { useEffect, useState } from 'react'
 import { Card, Form, InputGroup } from 'react-bootstrap'
 import { Typeahead } from 'react-bootstrap-typeahead'
 import { Trans, useTranslation } from 'react-i18next'
 
+import i18n from '../../i18n'
```
blanchardjeremy, line 8: I think we've been trying to not do relative imports. Could you make this an absolute import instead?

src/components/calculator/PrevalenceControls.tsx
```diff
@@ -139,6 +142,11 @@ export const PrevalenceControls: React.FunctionComponent<{
   setter: (newData: CalculatorData) => void
 }> = ({ data, setter }): React.ReactElement => {
   const { t } = useTranslation()
+  for (const iso_code of Object.keys(i18n.services.resourceStore.data)) {
+    countries.registerLocale(
+      require('i18n-iso-countries/langs/' + iso_code + '.json'), // eslint-disable-line @typescript-eslint/no-var-requires
```
blanchardjeremy, line 147: Curious for @beshaya's thought son this coding pattern.

src/components/calculator/PrevalenceControls.tsx
```diff
@@ -207,7 +215,16 @@ export const PrevalenceControls: React.FunctionComponent<{
...
+        const country_label =
+          Locations[locKey].iso3 &&
+          countries.getName(Locations[locKey].iso3!, i18n.language, {
+            select: 'official',
+          })
+            ? countries.getName(Locations[locKey].iso3!, i18n.language, {
+                select: 'official',
+              })
+            : Locations[locKey].label
+        return { label: country_label, value: locKey }
```
blanchardjeremy, line 227: image

Because we list states first and countries second, I don't know if it's immediately clear to a user that they can type country names in their language. I'm not sure what an easy way around this is. Maybe they just assume since everything else is translated, this will be translated too. But I think seeing "Alabama" as the first option might be a bit confusing.

The only slight improvement I can think of is to show more options in the typeahead by default, so they can scroll down and get a sense of the available options, and they'll soon realize they are translated.

Idea #2: We could also consider changing the placeholder text to have the name of a country where the currently selected language is commonly spoken. That way, they will be seeing it with an example value. Saying "select a country or US state" doesn't actually help that much, since the label above says the exact same thing. Placeholders are usually used as an example of what you could fill in.

Thoughts?

blanchardjeremy (changes requested):

src/components/calculator/PrevalenceControls.tsx
gyp: This is what https://github.com/michaelwittig/node-i18n-iso-countries suggests, as it does not load all countries by default in browser contexts to save bandwidth. If it needs to be changed, I'd need some help, I'm beyond my depth in React/TypeScript here.

src/components/calculator/PrevalenceControls.tsx
gyp: Sure, done.

src/components/calculator/PrevalenceControls.tsx
gyp: Yes, I also saw that to be the problem, though not a huge one.

One quick fix would be to move the US to be the first in the list or to move the US states after the countries. The country names  will be translated, so it will be evident that one needs to use localized names here. I'll try to add a patch that does that.

Idea no.2 is difficult for multiple reasons. For one, locales are about _languages_ and identified by ISO 639-1 two-letter codes. Locations are _countries_, identified by ISO 3166 codes. There's an overlap between these, but there's no 1:1 mapping and no reliable heuristic to convert between them (eg. should "Spanish" be converted to Spain or the country with the most Spanish speakers, or the most populous country that has Spanish as an official language...) Also, it's too easy to mistake the placeholder for the current selection and not understand why the calculator is telling you to select a location, even though one is already selected.

Prefilling using geolocation (#177) would probably be the best solution here.

gyp: Added a patch that moves the US states to the end:

image

It's a tradeoff, though: with this, we would be making it more difficult for US users, especially on mobile devices, to select their state as they need to type it in (or scroll endlessly to the bottom).

gyp: Hmm, looking at it even more I realized an even bigger problem: the options were sorted according to their English names rather than according to their name in the current locale. This could be extremely confusing (case in point: "États-Unis d'Amérique" appeared near Uruguay instead of near Estonia). I updated the code to sort the entries according to the locale. This solves the non-visible translation problem, too, but my previous comment about making it more difficult for US users still applies: it might even be more difficult at this point as states are intermingled between countries, making them harder to find.

blanchardjeremy: Hmmm. Dang. Yeah. I hear you on those problems. I'm not clear what the best solution is here. I do think our user base is US-centric, so maybe we just put it back to how it was (with your countries alphabetically sorted by locale-name at the bottom). I'm open to other solutions, but now that I'm seeing the other options, it seems like that's the least-bad option. What're your thoughts?

src/components/calculator/PrevalenceControls.tsx
blanchardjeremy: Okay. I'm good with leaving it as-is then! Let's call this point good.

src/components/calculator/PrevalenceControls.tsx
gyp: How about this: keep the original order with US states on the top when English language is selected and reorder alphabetically for other languages? We can assume that most non-English speakers come from outside the US, so US states are less relevant for them, so it's not a problem if they're a bit more difficult to find. The only possible caveat I see is with Spanish as that's also widely spoken in the US. We can make an exception for that as well and keep the US states on the top, too, but I'd vote against that: the potential Spanish user base is huge and probably bigger outside of the US.

I pushed a patch that makes this change, you can check it out in the Netlify preview.

blanchardjeremy (approved):

blanchardjeremy:
@gyp looks good. Will you resolve the merge conflict? Then go ahead and merge this into main.

PR #676, opened 2021-01-24
Obviate the need to clear the predefined box before selecting a new one.

beshaya:


blanchardjeremy:
I'm not clear what behavior is different with this pull request. Can you clarify so I know what I'm testing?

beshaya:
Before:
* Click premade activity
* enter an activity
* Click premade activity again
* Have to delete text or hit clear button to see options

After:
* click premade activity
* enter an activity
* Click premade activity again
* dropdown should have all entries.

blanchardjeremy:
Hmmm. I think it's a decent workaround given the situation.

Two requests to make it a more natural user interaction (so the user things the field is actually filled in, even though we're hacking it into placeholder):

* un-focus the field once an option is selected, so the cursor isn't sitting there flashing at the beginning of the field
* Make the placeholder text black instead of gray

The better approach would be to actually just find a way to change the default behavior to do what you want without hacking it into placeholder. But I'll settle for the above if that is easier for ya.

blanchardjeremy (dismissed):
Good start.

Two requests:

1. Can you make it so the cursor doesn't flash after an item is selected. It flashes at the beginning of the field, so that's confusing to the user, like they need to still type something even though they now see text.
2. If we're going to use the placeholder approach, can you make the placeholder text

I think this approach is a bit hacky. My ideal (that would solve the two above) would be to find a way to actually have the value be fully "set" into the field, and to adjust the behavior of the field to do what you want using some other means (hooks? idk.)

Those are my thoughts.

beshaya:
@blanchardjeremy @jy2wong how about this

> Good start.
> 
> Two requests:
> 
> 1. Can you make it so the cursor doesn't flash after an item is selected. It flashes at the beginning of the field, so that's confusing to the user, like they need to still type something even though they now see text.
> 2. If we're going to use the placeholder approach, can you make the placeholder text
> 
> I think this approach is a bit hacky. My ideal (that would solve the two above) would be to find a way to actually have the value be fully "set" into the field, and to adjust the behavior of the field to do what you want using some other means (hooks? idk.)
> 
> Those are my thoughts.

Try this new version?

blanchardjeremy:
Thanks for doing that. 

Could you get @jy2wong and @octern 's input on this before merging? They've been thinking a lot about this field. 

I'd love it if there was a smoother way that didn't require workarounds like turning it into a multiselect when it isn't really one. (By smoother I mean, the text is just black and filled in just like the state/region typeaheads).

I'd want their buy in before we finalize a solution here.

beshaya:
> Thanks for doing that.
> 
> Could you get @jy2wong and @octern 's input on this before merging? They've been thinking a lot about this field.
> 
> I'd love it if there was a smoother way that didn't require workarounds like turning it into a multiselect when it isn't really one. (By smoother I mean, the text is just black and filled in just like the state/region typeaheads).
> 
> I'd want their buy in before we finalize a solution here.

The correct thing might be to do the same thing for state/region as well. I don't think having black filled in text is actually desired since the field isn't a text field. If the text is black and filled in, clicking on it should let you edit the text (which is what multiselect=false does), but we explicitly don't want that. If I can't edit the text by clicking, it shouldn't look like that.

You could think of "multiselect" to just mean "i can select an option without clearing the selection first" - since the number of items that are selected concurrently is totally controllable outside the component. I don't think this is hacky at all.

The part of this I think is actually hacky is creating the fake entry to be filled in so that you can see / select the option that is already chosen. I'm ambivalent on this.

jy2wong:
This is not a solution so much as it is duct tape to hold the typeahead together until we can sort it out properly.

I don't think this is unambiguously an improvement though.

I also think I'd rather stick the old dropdown back in while we sort out the oddities of react-bootstrap-typeahead, because typeahead + bugginess + unintuitiveness is not, IMO, a clear improvement on "this dropdown is too long".

beshaya:
> This is not a solution so much as it is duct tape to hold the typeahead together until we can sort it out properly.
> 
> I don't think this is unambiguously an improvement though. It feels like clicking on the existing Token should actually do something. I don't think "you need to click in the whitespace" is very intuitive.
> 
> I also think I'd rather stick the old presets dropdown back in while we sort out the oddities of react-bootstrap-typeahead, because typeahead + bugginess + unintuitiveness is not, IMO, a clear improvement on "this dropdown is too long". (I think it'd be good to keep the country/county typeaheads, though.)

What behavior do you actually want? Can you point at another website that has it?

beshaya:
The token is now clickable. Do you also want to make it look like normal text (remove the different colored background, increase size, remove the 'x', etc)?

jy2wong:
The tokens behave as units, and typing backspace twice (one to "activate", one to actually remove) removes the entire preset. I think it might be confusing if it blended in with normal text. I think the current tag appearance is fine. Undecided on the X to clear. I do wish it was a bit bigger though; I think the typeahead multi makes the token's font size smaller to anticipate more selections (which isn't a concern here).

One thing I also like about the tokens is that the entire activity name is now visible on mobile (which was not true for the dropdown or single selection typeahead unless you had the menu open).

I like the current PR a lot better. Thumbsup on merging (possibly after increasing token font size?)

octern:
Agreed that making the text small enough for mobile is a win and that
pressing backspace twice to delete is unexpected, but not terrible. I don't
have a strong opinion overall and would definitely not block going forward
with it.

PR #681, opened 2021-01-25
Small miscellaneous UX tweaks in prevalence box and activity search

jy2wong:
* Add custom error message encouraging people to modify calculator
  values directly if there is no preset.
* Disable autofill addresses on the country, state and activity search
  typeaheads.
* Make the "switch to manual data entry mode" button smaller and boring
  (like an email unsubscribe button) so that people are less likely to
  hit it on autopilot
* Hide details if someone clicks "Switch to pre-populated location
  data". If someone is clicking on this button they probably want to
  reset back to the previous view.

### DIFF
--- a/src/components/calculator/PrevalenceControls.tsx
+++ b/src/components/calculator/PrevalenceControls.tsx
@@ -170,6 +170,7 @@ export const PrevalenceControls: React.FunctionComponent<{
 
   const handleSelectLocationButtonOnClick = () => {
     setLocationData('', '')
+    setDetailsOpen(false)
   }
 
   // If a stored location exists, load latest data for that location.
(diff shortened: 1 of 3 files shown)


blanchardjeremy:
for some reason, it needs to be:
`autocomplete="chrome-off"`
I don't know how it will work on other browsers, though.
`autocomplete="off"`  doesn't work.

Love all the other tweaks you made.

jy2wong:
type:search fixes the presets.
autoComplete:"off" fixes the country/state typeahead
autoComplete:"off" is not sufficient for the county typeahead, so I made it autoComplete:"chrome-off" as suggested. (???)

Tested in both firefox and chrome.

PR #684, opened 2021-01-26
Disable autocomplete in SavedDataSelector

jy2wong:


jy2wong:
@mattbell tested it, so gonna merge.

PR #685, opened 2021-01-26
Toggle prevalence manual data entry with a checkbox.

jy2wong:
* when toggling data entry mode, keep country/state/county information
  but mark it as disabled
  * previously, the button would jump away from user's cursor position,
    causing some confusion. Now it stays still.
  * previously, hitting the manual data entry button would clear
    existing location data. This isn't ideal for someone exploring the
    interface by clicking on everything.

### DIFF
--- a/src/components/calculator/PrevalenceControls.tsx
+++ b/src/components/calculator/PrevalenceControls.tsx
@@ -2,9 +2,10 @@ import i18n from 'i18n'
 import countries from 'i18n-iso-countries'
 import { isNumber } from 'lodash'
 import React, { useEffect, useState } from 'react'
-import { Card, Form, InputGroup } from 'react-bootstrap'
+import { Card, Form, InputGroup, ToggleButton } from 'react-bootstrap'
 import { Typeahead } from 'react-bootstrap-typeahead'
 import { Trans, useTranslation } from 'react-i18next'
+import { BsDash } from 'react-icons/bs'
 
 import ControlLabel from './controls/ControlLabel'
 import { ControlledExpandable } from 'components/Expandable'
@@ -13,7 +14,6 @@ import {
   calculateLocationPersonAverage,
   calculateLocationReportedPrevalence,
 } from 'data/calculate'
-import { TOP_LOCATION_MANUAL_ENTRY } from 'data/data'
 import { Locations, PrevalenceDataDate } from 'data/location'
 import 'components/calculator/styles/PrevalenceControls.scss'
 
@@ -30,8 +30,18 @@ const isTopLocation = (val: string): boolean => {
   return isFilled(val) && !!Locations[val]
 }
 
-const isManualEntry = (val: string): boolean => {
-  return val === TOP_LOCATION_MANUAL_ENTRY
+const PrevalenceFieldStatic: React.FunctionComponent<{
+  id: string
+  label: string
+  value: string | number
+  unit?: string
+}> = ({ id, label, value, unit }): React.ReactElement => {
+  return (
+    <div id={id}>
+      {label}: {typeof value === 'number' ? value.toLocaleString() : value}
+      {unit}
+    </div>
+  )
 }
 
 const PrevalenceField: React.FunctionComponent<{
@@ -41,7 +51,6 @@ const PrevalenceField: React.FunctionComponent<{
   unit?: string
   setter: (newValue: string) => void
   inputType: string
-  isEditable: boolean
   max?: number
   min?: number
   helpText?: string
@@ -53,26 +62,17 @@ const PrevalenceField: React.FunctionComponent<{
   setter,
   unit,
   inputType,
-  isEditable,
   max,
   min,
   helpText,
   className,
 }): React.ReactElement => {
-  if (!isEditable) {
-    return (
-      <div>
-        {label}: {typeof value === 'number' ? value.toLocaleString() : value}
-        {unit}
-      </div>
-    )
-  }
   let body: React.ReactElement = (
     <Form.Control
       className={'form-control form-control-lg col-md-3 col-lg-6 ' + className}
       type={inputType}
       value={value}
-      readOnly={!isEditable}
+      readOnly={false}
       onChange={(e) => {
         if (isNumber(max) || isNumber(min)) {
           let newValue = Number.parseFloat(e.target.value)
@@ -160,9 +160,6 @@ export const PrevalenceControls: React.FunctionComponent<{
   }
 
   const setLocationData = (topLocation: string, subLocation: string) => {
-    if (isManualEntry(topLocation)) {
-      setDetailsOpen(true)
-    }
     setter({
       ...data,
       ...dataForLocation(subLocation || topLocation),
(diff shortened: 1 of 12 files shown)


blanchardjeremy:
Omg I love this. So much cleaner. Thank you! 🙏

We just gotta get that failing test to pass. I'm actually not sure at first glance what's causing it. @beshaya ?

jy2wong:
Would like some coding feedback on that DONOTSUBMIT, which works around a type error I don't understand.
Also on the name useManualEntry for the new query param.

Would like some usability/testing feedback overall, too.

beshaya:
Oh, this is a bug. I didn't mean for prevalenceDataDate to be saved. Could
you remove it before storing in Calculator.tsx line 114? This could inflate
the prevalence calculated from % positive.

It makes sense that it becomes a string, since the code for persisting it
is JSON.stringify(...)

blanchardjeremy:
I'm excited for this to launch whenever it's ready!

blanchardjeremy:
I think @beshaya would be the best person to review this since I think he might have some advice on some of the coding patterns.

Functionally, it looks great!

src/components/Expandable.tsx
```diff
@@ -5,6 +5,7 @@ import 'components/styles/Expandable.scss'
 
 export const ControlledExpandable: React.FunctionComponent<{
   id: string
+  noContract?: boolean
```
beshaya, line 8: Perhaps this should be "clickable"? 

Prefer using boolean to optional<boolean>. Optional<boolean> implies three states, and doesn't imply what the default behavior is.

beshaya: A more common/utilitarian solution might be to optionally take an onClick function which overrides the standard onClick.

src/components/calculator/PrevalenceControls.tsx
```diff
@@ -306,31 +330,28 @@ export const PrevalenceControls: React.FunctionComponent<{
...
-        </span>
-      )}
+          </span>
+        </ToggleButton>
+      </span>
       <ControlledExpandable
         id="prevalence-details"
         header={t('calculator.prevalence.details_header')}
-        headerClassName={isManualEntryCurrently ? 'd-none' : ''}
+        noContract={isManualEntryCurrently}
```
beshaya, line 354: Is this field being added because the manual entry is actually inside the expandable?

That seems odd to me. I think this would be easier to reason about this if a different component that gets swapped in rather than having a complicated override of how an expandable works.

src/pages/Calculator.tsx
```diff
@@ -107,13 +107,21 @@ export const Calculator = (): React.ReactElement => {
       recordCalculatorChanged(expectedValue)
     }
 
+    const removePropsForLocalStorage = (incomingData: CalculatorData) => {
```
beshaya, line 110: perhaps easier to read as storeDataLocally() which also stores the data?

src/pages/Calculator.tsx
```diff
@@ -123,9 +131,11 @@ export const Calculator = (): React.ReactElement => {
     return { points: expectedValue, lowerBound, upperBound }
   }, [calculatorData, setQuery])
 
+  // Location mode and not blank OR fully valid manual mode
```
beshaya, line 134: Could you either make the two sides of this OR statement follow the same pattern or remove it?

// Location mode and location chosen OR manual mode and filled out.

src/data/calculate.ts
```diff
@@ -191,6 +193,9 @@ export const calculateLocationPersonAverage = (
 
     return personRisk * ONE_MILLION
   } catch (e) {
+    console.log(
```
beshaya, line 196: Remove

src/data/calculate.ts
```diff
@@ -82,7 +84,7 @@ export const DAY_0 = new Date(2020, 1, 12)
 const MS_PER_DAY = 1000 * 60 * 60 * 24
 
 // From https://covid19-projections.com/estimating-true-infections-revisited/
-const prevalanceRatio = (positivityPercent: number | null, date: Date) => {
+const prevalenceRatio = (positivityPercent: number | null, date: Date) => {
```
beshaya, line 87: Thanks!

src/components/Expandable.tsx
beshaya, line 8: IIUC, the way you are using this is to make an object that normally can be opened or closed via clicking not work. 

I think it is better for user expectations if all icons always work a certain way (i.e. all chevrons are clickable); could we use a different component for "thing that is expanded and cannot be clicked to collapse?"

beshaya: Oh, also this leaves the chevron + header with a different cursor when mousing over, which is misleading.

src/pages/Calculator.tsx
jy2wong: removed

src/components/calculator/PrevalenceControls.tsx
jy2wong: the manual entry swap was done at the PrevalenceField level. If isManualEntryCurrently was true, it would flip from the form input to the "statement" form. The tradeoff is a bit more mess in ControlledExpandable to get the smooth animation working between manual entry and details. Getting that animation working properly without having it all be in the same <Collapse> is really tricky.

See the WIP commit, which cleans up ControlledExpandable (and some of PrevalenceControls) but removes the animation between manual and location mode.

src/components/Expandable.tsx
jy2wong: Better now? Made the icon flat when it's not contractable.

Also separated out the stuff that shows up in manual entry mode vs regular old details and as a happy side effect got rid of the "remember last location mode details open state" logic.

src/components/calculator/PrevalenceControls.tsx
```diff
@@ -204,9 +216,11 @@ export const PrevalenceControls: React.FunctionComponent<{
     isTopLocation(data.topLocation) &&
     Locations[data.topLocation].subdivisions.length > 1
 
-  const locationSet = isTopLocation(data.topLocation)
+  const locationSet = !data.useManualEntry && isTopLocation(data.topLocation)
 
-  const isManualEntryCurrently = isManualEntry(data.topLocation)
+  const [isManualEntryCurrently, setIsManualEntryCurrently] = useState(
```
beshaya, line 221: Nit: I think this is less likely to get accidentally broken in the future if you specify the type of the generic: `useState<boolean>`

src/components/calculator/PrevalenceControls.tsx
```diff
@@ -256,16 +270,17 @@ export const PrevalenceControls: React.FunctionComponent<{
...
           id="top-location-typeahead"
           header={t('calculator.select_location_label')}
         />
         <Typeahead
           clearButton={true}
+          disabled={isManualEntryCurrently}
           highlightOnlyResult={true}
           id="top-location-typeahead"
-          inputProps={{ autoComplete: 'off' }}
+          inputProps={{ autoComplete: 'chrome-off' }}
```
beshaya, line 283: Just checking, does this stop autocomplete on other browsers?

beshaya (approved):
This looks really good. Left a few questions and a request for a refactor (deferrable), but nothing blocking.

src/components/calculator/PrevalenceControls.tsx
jy2wong: I checked firefox and chrome, but not any others.

jy2wong:
hey Ben, could you explain the refactor request in more detail (or point me to a comment about a refactor that I missed?)

src/components/calculator/PrevalenceControls.tsx
```diff
@@ -306,59 +322,48 @@ export const PrevalenceControls: React.FunctionComponent<{
...
-          isEditable={isManualEntryCurrently}
-          className="hide-number-buttons"
-        />
-        {locationSet && data.casesIncreasingPercentage === 0 ? (
-          <div>{t('calculator.prevalence.cases_stable_or_decreasing')}</div>
-        ) : (
+          </span>
+        </ToggleButton>
+      </span>
+      {isManualEntryCurrently ? (
```
beshaya, line 343: The two chunks that get swapped between here seem like a good candidate for refactoring into their own components. Possibly put them in new files, components/calculator/prevalence/LocationPrevalenceDetails.tsx and components/calculator/prevalence/ManualPrevalenceDetails.tsx

Could be deferred to a later PR.

beshaya (commented):
This comment apparently got eaten by github on my last review 🤷

src/components/calculator/PrevalenceControls.tsx
jy2wong: done

blanchardjeremy:
Is this good to go? Or should we have @beshaya give it another review?

blanchardjeremy:
Oh I guess @beshaya approved it with his last review. So I say merge away! Thanks for doing this!

PR #689, opened 2021-01-27
Cleanup front page intro text

blanchardjeremy:
Based on user testing feedback:

> UX Interview observation: Users were frequently derailed by the "Model Update" box at the top of the page. One of them said she didn't know what it meant, so she assumed the site wasn't for people like her. Another one did know what it meant but still saw it as being so technically detailed that she assumed the site wasn't for her.
> Could we consider moving updates and non-essential information to a sidebar? (several users also mentioned the conspicuously empty space on the right side of the page)

This is my first take on it. @octern is going to draft up something that he put more thought into and we can use the PR to implement whatever he/we comes up with.

### DIFF
--- a/src/components/calculator/FirstTimeUserIntroduction.tsx
+++ b/src/components/calculator/FirstTimeUserIntroduction.tsx
@@ -1,7 +1,5 @@
 import React from 'react'
-import { Alert } from 'react-bootstrap'
 import { Trans, useTranslation } from 'react-i18next'
-import { BsCheckBox } from 'react-icons/bs'
 
 import { Expandable } from 'components/Expandable'
 
@@ -12,6 +10,7 @@ export function FirstTimeUserIntroduction(): React.ReactElement {
       <Expandable
         id="first-time-user-introduction"
         header={t('calculator.firsttime.microcovid_header')}
+        open={true}
       >
         <Trans>calculator.firsttime.microcovid_explanation</Trans>
       </Expandable>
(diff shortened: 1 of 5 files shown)


octern:
I'm imagining the sidebar could ultimately include links to blog posts and maybe a sticky post at the top with the instructional video Finan recorded. It would need to be able to scroll vertically, independently of the rest of the page. Is that a possibility?

blanchardjeremy:
@octern that's definitely possible.

I've lost track of where we're at with proposed solutions in Slack for this.

I'm currently holding this PR as "designers, please tell us what you want, and we'll do it."

Please let me/us know when we're to that point. :)

jy2wong:
Observations: in the wireframe, there was plenty of room for two news notifications, but this somehow disappeared on the actual site. I'll probably just remove the b117 update the next time I push to github, but wanted to point it out this big blank space as another place that a video tutorial thumbnail could go.

Other observations: because we require location to be filled out before any of the activity/relationship card shows up, clicking on the example activity links doesn't visibly do anything on the page (not that I've figured out yet how to select a preset on click, but...)

I proposed splitting the first section of the calculator into its own card with a colourful heading a couple days ago; I think we'd need something like this in before making the example scenarios clickable.

I suggest, as far as this particular wireframe is concerned:
1. That we check in a slight variant on the wireframe in which the examples are plain text, not links, very soon (for translation reasons, I'll put span tags where the links ought to go)
2. I'll try to pull out the preset selection dropdown so that it's always visible even if location hasn't been set
3. As soon as the preset selection dropdown is always visible, I'll turn the examples into links (I'll probably have some questions for @beshaya or @blanchardjeremy about how to select activity preset on link).

PR #690, opened 2021-01-27
Switch mask info section on explanation card to use our existing component

blanchardjeremy:
From a report from @octern 

> Probable bug: the "learn more about masks" collapsible doesn't cause a cursor change on hover.


beshaya (approved):
LGTM

PR #695, opened 2021-01-28
Update prevalence 2021-01-28

blanchardjeremy:


blanchardjeremy:
Uh oh. The country list isn't properly sorted. It's mixed in with states. 

This seems to be an artifact from #675.

@gyp can you take a look and let me know why they aren't sorting properly?


image

gyp:
Hmm. A switch to a different language and then back fixes it. I won't be able to look at it before tomorrow, but you should check around this if you have some time:

https://github.com/microcovid/microcovid/blob/89b4001e0984e6a7d57a10cbeb92bb3ebde39979/src/components/calculator/PrevalenceControls.tsx#L234

`language` is probably not set if at least one switch is not made with the selector. I think we could safely check for it being null and skip the sort in that case, too.

blanchardjeremy:
I see. Okay I'll merge this since it's not a bug related to this PR.

PR #699, opened 2021-01-29
Add fixed to make bold text render properly from google font and darker overall

blanchardjeremy:


beshaya:
I think @gyp replaced Lato with LatoLocal because Lato is broken for some non english languages. Let's make sure we don't break our friends across the ocean!

gyp (changes requested):
Yep, that is correct, that's why I did that, see 6b75cd70  I'd prefer to keep it if possible, but, truth be told, the difference is minor and mostly only annoys me out of pedantry. I'd rather we kept it, but I can live with reverting back to the older font version that's available in Google Fonts. If we do that, however, we should remove the custom fonts I added under src/fonts.

blanchardjeremy:
Yeah, I think we just need to add the font files for bold and it should work. I'll work on it this weekend.

blanchardjeremy:
Replaced by #734

PR #712, opened 2021-02-01
[inactive] Move risk tracker dropdown out of nav bar.

jy2wong:


jy2wong:
ugh, the text is misaligned. Will fix. How's this functionally?

I have "manual" in right now, but any thoughts on the following alternatives?
* Guide
* Documentation
* Jump to...

jy2wong:
I guess "instructions" could also work?

beshaya:
Thanks for putting this together. Two things that jump out to me:

* The menu is where the "previous page" link would be, which seems not very intuitive to me.
* The menu is unreachable from any of the other pages, which doesn't seem ideal.

Perhaps putting the menu in the middle, between the previous and next links would work better?

Also, possibly "Jump to..." would be a more action-oriented label for the menu?

Also, possibly I have no idea what I'm talking about, I'm not even sort of an average user :)

blanchardjeremy:
@beshaya basically said what I was going to say. I have those same initial reactions. Seems odd to make it hard for people to understand the overall structure.

What would you think about having a "secondary navigation bar" that lives under the title that has all the subpages (since there are only a few) listed a horizontal menu, with the current page marked as "active"?

blanchardjeremy:
Closing since we haven't touched it in a while!

PR #718, opened 2021-02-02
Remove healthcare worker category / intermediate method

beshaya:
Includes detailed arguments in the Q&A for why this category doesn't make sense.


catherio:
this is extremely clear; the multiplied-through example drives the point home very effectively; the wording is straightforward; explaining why we used to do it a different way is A+ on transparency; overall very enthusiastic about this and getting this out quickly!

beshaya:
Yes! The broad risk categories were never meant for your friends!

PR #727, opened 2021-02-06
"Korea, South" -> "South Korea"

jy2wong:


blanchardjeremy (approved):
Thank you SO much for fixing this. It's always bothered me, I just haven't cared enough to do anything about. 

LGTM!

PR #734, opened 2021-02-08
Fix bold fonts to use proper bold instead of faux bold

### LINKED ISSUE #735: Fonts are using "faux bold" instead of proper bold
Faux bold makes things look fuzzy and not-so-great.

Currently looks like
image

Should look like:
image

(although I'm also making the text black instead of dark grey in the second version, so you're seeing two changes at once, but you get the idea)

blanchardjeremy:
Closes #735


src/styles/App.scss
```diff
@@ -56,6 +88,10 @@ h1,h2,h3,h4,h5,h6 {
   margin-bottom: 1.3rem;
 }
 
+b, strong {
+  color: black; // Turn dark-gray text black to make bold stand out more
```
jy2wong, line 92: Looks great in the blog, risk tracker docs and whitepaper!

It also seems to affect many places where it's not quite appropriate on the calculator itself though (model update; form labels now darker than e.g. the Precaution subheading etc)

is there a way to be more selective about which bold/strong text gets turned to black?

jy2wong (changes requested):

src/styles/App.scss
blanchardjeremy: Sure! Done.

jy2wong (approved):
looks good to me!

PR #768, opened 2021-02-21
ben/vaccine ux

beshaya:
- Add yourVaccine to Calculator core
- Fix typo
- fix typo
- Add vaccine ux


src/locales/en.json
```diff
@@ -31,7 +31,13 @@
       "their_mask_header": "Their mask",
       "filtered_ensure_airflow": "Ensure sufficient flow through the filter for your room. For help on how to do this, see <1>here</1>.",
       "volume_header": "Volume",
-      "volume_question": "How much is everyone talking?"
+      "volume_question": "How much is everyone talking?",
+      "your_vaccine_header": "Your Vaccine",
+      "your_vaccine_label": "Which vaccine did you receive?",
+      "your_vaccine_doses_label": "How many dose have you received?",
+      "no_vaccine": "No vaccine",
+      "doses_7_days_ago": "Only count a dose if you received it at least 7 days ago.",
```
jy2wong, line 39: "If your last dose was <1 week ago, subtract one dose when answering--the last dose may not be protecting you yet."

is too long, but addresses the following:

1. With the current version, I have to read the entire line before I can decide if it applies to me
2. If I match the condition I pattern-matched from this instruction (at least 7 days ago), I need to do nothing. In fact, if I was last shot *at most* 7 days ago, I need to take extra action. This is an opportunity for error.
2. If I received my last dose 3 days ago and read the entire line and decided it applied to me, I have to then sort out what it means to count a dose and what it means to *not* count a dose. With some thought, I might come to the conclusion that I need to subtract one from the number I started with and then separately perform the subtraction, or do this: "I had two shots. Does the first one count? Yes. Does the second one count? No." This is more effort than strictly necessary.

src/locales/en.json
```diff
@@ -232,6 +238,12 @@
       "microcovid_budget_one_percent": "Keeps to a budget of 200 microCOVID per week (\"Standard Caution Budget\")",
       "microcovid_budget_deci_percent": "Keeps to a budget of 20 microCOVID per week (\"High Caution Budget\")",
       "hasCovid": "Has COVID"
+    },
+    "vaccine": {
+      "pfizer": "Pfizer BioNTech",
+      "moderna": "Moderna",
+      "astra_zeneca": "AstraZeneca",
+      "unknown": "Unknown or other"
```
jy2wong, line 246: I'm not sure how well this will be received given the experience with "Other" and masks, but I also don't have a confident suggestion. @octern ?

src/posts/paper/99-changelog.ts
```diff
@@ -7,6 +7,12 @@ interface Change {
 }
 
 const changes: Change[] = [
+  {
+    date: new Date(2021, 1, 21),
+    content: `
+* Added precaution for being vaccinated. See paper [Q&A](/paper/13-1-and-a#vaccines-qa) and [Research Sources](/paper/14-research-sources#vaccines) for details.
```
jy2wong, line 13: This should probably exist on the front page too. "Added 'your vaccine' precaution" for brevity?

src/locales/en.json
beshaya: With masks, I know what mask I have and it's not on the list. With vaccines, it's much more likely that a person just doesn't know which one they got.

src/posts/paper/99-changelog.ts
beshaya: Ah, I see someone removed the code that puts the changelog on the front page. Grumbles. Done.

src/locales/en.json
beshaya: Done.

I honestly don't care that much about errors here. It'll only affect people for a week, and the actual # of days you need to wait is actually fuzzy; they had to choose an endpoint before starting the studies. Plausibly you're protected by the time your fever from the second dose stops. I'm not terribly worried about people with fevers, chills, and aches from the second dose getting into trouble, and the multiplier from the first dose is so mild that I don't think anyone will get into much trouble. Erroneously doubling your budget for a week is negligible on a year scale.

src/locales/en.json
```diff
@@ -166,6 +172,7 @@
     "risk_group_empty_warning": "Enter your location first.",
     "saved_scenario_loaded_message": "If needed, adjust the following prefilled answers to match your <strong>{{ scenarioName }}</strong> scenario.",
     "alerts": {
+      "vaccines_update": "<0> February 23rd, 2021:</0> We added a precaution section for being vaccinated. Learn more in the <2>Q&A</2>.",
```
jy2wong, line 175: I think this is too jargon-y. If it was my first time visiting the calculator, I would have no reason to know what a precaution section is. Thoughts on:

* We added a "your vaccine" precaution.
* The calculator now gives a bonus if you are vaccinated.

beshaya:
Either is fine. Separately, the changelog is really specifically for
returning users. Should we figure out how to target it at them better?

src/locales/en.json
beshaya: Done.

beshaya: Done.

src/pages/Calculator.tsx
```diff
@@ -341,6 +339,28 @@ export const Calculator = (): React.ReactElement => {
                       repeatedEvent={repeatedEvent}
                     />
                   </Col>
+                  <Col xs="12" id="vacines" className="calculator-params">
```
jy2wong, line 342: nit: vacines -> vac**c**ines

beshaya: done.

jy2wong (approved):
lgtm other than those small nits

PR #772, opened 2021-02-23
Fix percentage chance for 20 microcovids in calculator explainer

jy2wong:


jy2wong:
I checked the other numbers in the paragraph too; this is the only one that needs fixing.

jy2wong:
(since this is a 2 character deletion typo fix in a translation string and the original intent is very clear from context, I'm just going to merge it in without going through review.)

PR #779, opened 2021-03-02
Kiss multiplier

jfslocum:
Updates kiss multiplier to 5x (in line with singing/heavy exertion). Drops minimum time floor to 20 minutes - "minimum exposure risk" then stays roughly the same.


src/data/__tests__/calculate.test.ts
```diff
@@ -373,22 +373,24 @@ describe('calculate', () => {
       expect(calcValue(unmaskedIntimate)).toEqual(calcValue(maskedIntimate))
     })
     it('should be at least 12% (1 hr) transfer risk.', () => {
```
beshaya, line 375: nit: update this test case name?

src/data/__tests__/calculate.test.ts
```diff
@@ -373,22 +373,24 @@ describe('calculate', () => {
...
       }
       const oneMinuteIntimate: CalculatorData = {
-        ...oneHourIntimate,
+        ...twentyMinuteIntimate,
         duration: 1,
       }
-      const twoHourIntimate: CalculatorData = {
-        ...oneHourIntimate,
-        duration: 120,
+      const thirtyMinuteIntimate: CalculatorData = {
```
beshaya, line 384: Could you also add a case that checks that the correct cap is applied?

beshaya (approved):
LGTM % nit + request for test case.

PR #793, opened 2021-03-10
Add risk tracker version number so it can be imported to each copy of the spreadsheet

blanchardjeremy:


blanchardjeremy:
I'm just gonna merge it.

PR #821, opened 2021-04-02
Easier Risk Tracker release process

blanchardjeremy:
Move the version number next to the spreadsheet URL so we can change everything at once.


beshaya:
LGTM. It would be even better if there were a way to get it from the /public folder, since now the version # exists in two places.

beshaya (approved):

PR #839, opened 2021-04-28
Update prevalence 2021-04-28

gyp:


gyp:
@blanchardjeremy I remember that you've been doing this in the past but now it's been a while since the last update and things changed dramatically in the last week where I live so I wanted to put an update in. Let me know if you'd like to keep doing these or you'd prefer if I took this one over. This part is Python, which is closer to my heart than NodeJS, so I'd be happy to keep the fetching up-to-date in case there are any breaking API changes.

PR #858, opened 2021-05-25
Update single shot efficacy of Moderna/Pfizer

melonmouse:
The CDC reports an efficacy of 80%, 14 days after a first shot of Moderna or Pfizer. This figure includes both symptomatic and asymptomatic cases:
https://www.cdc.gov/mmwr/volumes/70/wr/mm7013e3.htm?s_cid=mm7013e3_w

When more data was collected, the figure was updated to 82%: https://www.cdc.gov/media/releases/2021/p0514-covid-19-vaccine-effectiveness.html

To me, based on these results, it would seem more reasonable to assume that asymptomatic cases are prevented at the same rate that symptomatic cases are prevented. If so, the (safe/pessimistic) correction for asymptomatic cases, as used in the current figure ((0.4 + .42 * .23) / (0.8 + .42 * .2) = 0.56), could be omitted. The figures for the efficacy of other vaccines could be updated based on that.

Discussion:
- this data was gathered among health care workers only (although they did partial corrections for this)
- the 95% CI reported by CDC is 80% [59%–90%]
- the CDC warns that sensitivity of the used PCR tests could affect the outcome of the study

Is 56% still the best estimate? Is 20% more reasonable? Should we go for a middle ground? Let's discuss.

p.s. I love the microcovid project, thanks for working on this!!!

### DIFF
--- a/src/data/data.ts
+++ b/src/data/data.ts
@@ -364,11 +364,11 @@ export interface VaccineValue {
 export const Vaccines: { [key: string]: VaccineValue } = {
   pfizer: {
     label: i18n.t('data.vaccine.pfizer'),
-    multiplierPerDose: [1, 0.56, 0.1],
+    multiplierPerDose: [1, 0.2, 0.1],
   },
   moderna: {
     label: i18n.t('data.vaccine.moderna'),
-    multiplierPerDose: [1, 0.56, 0.1],
+    multiplierPerDose: [1, 0.2, 0.1],
   },
   astraZeneca: {
     label: i18n.t('data.vaccine.astra_zeneca'),


beshaya:
Thanks for looking into this.

I have a hard time reconciling the CDC's study, which states 80% reduction, N=5000 with [Israel's study](https://www.nejm.org/doi/full/10.1056/NEJMoa2101765), which found 50% reduction for asymptomatic + symptomatic and 60% reduction for symptomatic cases for days 14-21, N=180,000. 

Do you see flaws in one of the two studies? If not, I'm inclined to stick with numbers from the larger study.

jfslocum:
I would also point out that the CI for the CDC study contains the numbers found by the Israeli study, so they don't even really contradict each other. There's also a UK study, which is in between the CDC and Israeli studies both in terms of number of participants (~25k) and efficacy found (~72%) - this study's CI *also* contains the results of the Israeli study. 

The only issue with the Israeli study that I know of is they don't measure asymptomatic efficacy very well.

melonmouse:
Thanks for your replies!

I wasn't aware that the Israel study had a much larger sample size (and a much smaller CI). It does look like 50% is the better estimate, and probably the best estimate we can use right now. Closing this pull request.

An issue with all of these studies is that they do not correct for behavioral changes induced by receiving the vaccination (the control group did not receive a placebo). If we are to expect that people get less careful once vaccinated, the estimated efficacy could be on the low side. In the Israeli study, they do not analyze the period in which this would be most apparent: "The period immediately after the first dose, when immunity is gradually building, was excluded in the main analyses because the risk ratio is expected to be close to 1 during this period." - the risk ratio may be >1 during that period (as people have more contact, but the vaccination doesn't help yet).

PR #860, opened 2021-05-28
Make update_prevalence more robust to negative daily cases

beshaya:
The JHU dataset is not strictly increasing case numbers (perhaps due to governments
correcting their numbers?). If a correction happens, compute # of cases in a week as
max(cumulative_cases_past_week) - min(cumulative_cases_past_week)

It is difficult to know if this is over or underestimating the actual number of cases,
since we have no way to determine the relative magnitudes of the correction vs the #
of new cases.

### DIFF
--- a/update_prevalence.py
+++ b/update_prevalence.py
@@ -193,17 +193,17 @@ class Place(pydantic.BaseModel):
     tests_in_past_week: Optional[int]
 
     @property
-    def recent_daily_cases(self) -> List[int]:
+    def recent_daily_cumulative_cases(self) -> List[int]:
         """Returns a list whose last entry is the most recent day's
-        case count, and earlier entries are earlier days' counts.
-        So recent_daily_cases[-5] is the number of cases reported
-        5 days ago.
+        cumulative case count, and earlier entries are earlier days' counts.
+        So recent_daily_cumulative_cases[-5] is the total number of cases reported
+        up to 5 days ago.
         """
-        daily_cases = []
+        daily_cumulative_cases = []
         current = effective_date
         if current not in self.cumulative_cases:
             raise ValueError(f"Missing data for {self.fullname} on {current:%Y-%m-%d}")
-        while len(daily_cases) < 14:
+        while len(daily_cumulative_cases) < 14:
             prev = current - timedelta(days=1)
             if prev not in self.cumulative_cases:
                 if prev > min(self.cumulative_cases.keys()):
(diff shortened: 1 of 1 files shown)


oremanj (approved):

update_prevalence.py
```diff
@@ -193,17 +193,17 @@ class Place(pydantic.BaseModel):
...
     @property
-    def recent_daily_cases(self) -> List[int]:
+    def recent_daily_cumulative_cases(self) -> List[int]:
         """Returns a list whose last entry is the most recent day's
-        case count, and earlier entries are earlier days' counts.
-        So recent_daily_cases[-5] is the number of cases reported
-        5 days ago.
+        cumulative case count, and earlier entries are earlier days' counts.
+        So recent_daily_cumulative_cases[-5] is the total number of cases reported
+        up to 5 days ago.
```
jfslocum, line 200: Nit: "as of 5 days ago."

update_prevalence.py
```diff
@@ -214,15 +214,21 @@ def recent_daily_cases(self) -> List[int]:
...
     @property
     def cases_last_week(self) -> int:
-        return sum(self.recent_daily_cases[-7:])
+        return max(self.recent_daily_cumulative_cases[-7:]) - \
+            min(self.recent_daily_cumulative_cases[-7:])
+
+    @property
+    def cases_week_before(self) -> int:
+        return max(self.recent_daily_cumulative_cases[-14:-7]) - \
+            min(self.recent_daily_cumulative_cases[-14:-7])
```
jfslocum, line 231: Maybe abstract this and previous method as "cases_n_weeks_prior"?

update_prevalence.py
```diff
@@ -214,15 +214,21 @@ def recent_daily_cases(self) -> List[int]:
...
+                self.cumulative_cases[current]
             )
             current = prev
-        return daily_cases[::-1]
+        return daily_cumulative_cases[::-1]
 
     @property
     def cases_last_week(self) -> int:
-        return sum(self.recent_daily_cases[-7:])
+        return max(self.recent_daily_cumulative_cases[-7:]) - \
```
jfslocum, line 225: I'd like to see a comment explaning why max-min makes sense. Or, alternatively, why the previous behavior was leading to bugs.

update_prevalence.py
```diff
@@ -214,15 +214,21 @@ def recent_daily_cases(self) -> List[int]:
...
+            min(self.recent_daily_cumulative_cases[-7:])
```
jfslocum, line 226: This should be `max(self.recent_daily_cumulative_cases[-8:]) - min(self.recent_daily_cumulative_cases[-8:])` - *usually* that will work out to the same as  `self.recent_daily_cumulative_cases[-1:] - self.recent_daily_cumulative_cases[-8:]`

PR #861, opened 2021-05-28
Update true infections formula

beshaya:
- Update prevalence calculation based on 10 December 2020 Covidestim model.
- Update prevalence calculation based on COVID19 Projections' December 10 update

### DIFF
--- a/src/data/__tests__/calculate.test.ts
+++ b/src/data/__tests__/calculate.test.ts
@@ -39,7 +39,8 @@ const baseTestData = {
   casesPastWeek: 2999, // will add 1 in pseudocount
   casesIncreasingPercentage: 0,
   positiveCasePercentage: 0,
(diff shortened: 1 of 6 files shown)


blanchardjeremy (approved):
LGTM

blanchardjeremy:
FYI - relevant page - https://covid19-projections.com/estimating-true-infections-revisited/

blanchardjeremy:
Question: Where did you get the 30 day delay (instead of 25?)

beshaya:
For the test? The old formula was 1250 / (25 + n), and I had n set to 25.
1250 / (25 + 25) = 25. The new formula is 1000 / (10 + n). 1000 / (10 + 30)
= 25.

Easier to make one change in the test than try to track down everywhere the
prevalence would change of the projection formula changing.

PR #866, opened 2021-06-01
Add vaccine data to data injest.

beshaya:
- Loads vaccine doses + types for US States from JHU
- Loads vaccine doses for Countries from JHU
- Loads vaccine doses for Counties from CAN
- Applies the ratio of vaccine types from states to counties
- Estimates the prevalence among unvaccinated people (used to compute vaccinated prevalence)

This assumes that countries w/o vaccine type data are getting AZ,
which results in under-estimating the risk of random vaccinated
or unvaccinated people in e.g. Israel.


blanchardjeremy:
@beshaya - I haven't looked at the code yet, but I just had the thought: I wonder how we indicate to the user that "average local person" does in fact account for the amount of people who are vaccinated. --- I'm worried folks night see "average person in my area" and have a thought "but wait half of the people in my area are vaccinated! Does this number account for that?!" --- we know it does. But they don't because it's not stated. 

This is a minor point and happy if we don't do anything about it. But wanted to bring it up.

beshaya:
Yeah, I think this is a point of confusion. Do you have any ideas on how to
signal this?

blanchardjeremy:
@beshaya simplest solution I can think of is a notice below the risk profile section if they select "average person" that explains how the average automatically accounts for vaxxed and unvaxxed.

update_prevalence.py
```diff
@@ -18,6 +19,25 @@
 from typing import Optional, ClassVar, Iterator, List, Dict, Type, TypeVar, Any
 from us_state_abbrev import us_state_name_by_abbrev
 
+VACCINE_MULTIPLIERS = {
```
jfslocum, line 22: Is there a way to reference the existing vaccine multiplier constants, instead of duplicating them here? I'm sure this list will change at some point - best to have a single copy so we don't have to track them all down for updating.

update_prevalence.py
```diff
@@ -235,6 +288,66 @@ def cases_week_before(self) -> int:
...
+                self.vaccines_total.completed_vaccinations + 
+                self.vaccines_total.partial_vaccinations
+            )
+
+        assert (
+            total_vaccinated <= self.population
+        ), f"{self.label} has over 100% vaccinated. Population: {self.population}, vaccinations: {total_vaccinated}"
+
+        if total_vaccinated == 0:
+            return
```
jfslocum, line 345: Shouldn't this case just return 1.0?

update_prevalence.py
```diff
@@ -235,6 +288,66 @@ def cases_week_before(self) -> int:
...
+    def partial_vaccination_total(self):
+        if (self.vaccines_by_type is not None):
+            return reduce(lambda x, key: x + self.vaccines_by_type[key].partial_vaccinations, self.vaccines_by_type, 0)
+        return self.vaccines_total.partial_vaccinations
+
+    # Compute the estimated risk ratio for an unvaccinated person vs an average person.
+    # Average prevalence = sum(prevalence in group * proportion of population in group)
+    #                    = sum(vaccine_mult * unvaccinated_prev * proportion of population)
+    # Unvaccinated risk / Average risk = sum(vaccine_mult * proportion of pop)
+    #                                  = population / sum(vaccine_mult * # of vaccinations)
```
jfslocum, line 317: I think this should be `sum(vaccine_mult * # of vaccinations) + num_unvaccinated`

update_prevalence.py
```diff
@@ -235,6 +288,66 @@ def cases_week_before(self) -> int:
...
+        self.vaccines_total.partial_vaccinations = partial_vaccinations
+        self.vaccines_total.completed_vaccinations = complete_vaccinations
+    
+    def set_vaccines_of_type(self, vaccine_type: str, partial: int, complete: int):
+        if self.vaccines_by_type is None:
+            self.vaccines_by_type = {}
+        
+        self.vaccines_by_type[vaccine_type] = Vaccination()
+        self.vaccines_by_type[vaccine_type].partial_vaccinations = partial
+        self.vaccines_by_type[vaccine_type].completed_vaccinations = complete
```
jfslocum, line 301: I think these should unset the other: `set_total_vaccines` should clear `vaccines_by_type` and `set_vaccines_of_type` should clear `vaccines_total`.

update_prevalence.py
beshaya: Yeah! I can read them from the csv file.

We still have 2 copies, but at least it's not more.

update_prevalence.py
beshaya: I interpret 0 vaccines as not having data, which means we should refuse to compute "average vaccinated" This will propagate through as a null field in the data export.

update_prevalence.py
beshaya: It is; I consider "unvaccinated" to be a vaccination state. Modified to be clearer.

update_prevalence.py
beshaya: That would not work as intended - CAN and JHU both provide state data, but CAN only provides total vaccines while JHU provides vaccine breakdowns. We injest JHU first, so we don't want to override that when adding CAN.

I'd rather just store both and then use the best data when processing. Left a comment on the fields.

PR #875, opened 2021-06-14
Add theirVaccine to calculator core.

beshaya:
Only affects "average" risk profiles. More thought needed for supporting
calculated risk profiles.

### DIFF
--- a/src/data/calculate.ts
+++ b/src/data/calculate.ts
@@ -33,6 +33,9 @@ export interface CalculatorData {
   casesIncreasingPercentage: number
   positiveCasePercentage: number | null
   prevalanceDataDate: Date
+  completeVaccinations: number | null
+  unvaccinatedPrevalenceRatio: number | null
+  averageFullyVaccinatedMultiplier: number | null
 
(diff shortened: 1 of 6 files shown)


src/data/calculate.ts
```diff
@@ -237,16 +243,42 @@ export const calculatePersonRiskEach = (
...
       })
-    )
+
+    if (
+      // TODO: Support vaccinated risk for other Risk Profiles.
+      data.riskProfile !== 'average' ||
+      data.unvaccinatedPrevalenceRatio === null ||
+      data.averageFullyVaccinatedMultiplier === null
+    ) {
+      return unadjustedRisk
```
blanchardjeremy, line 263: Obviously this is beyond the scope of this PR, but we should make sure to indicate to people whether or not we're accounting for their vaccine. (Or only allow them to select it if we can account for it. And if we can't yet account for it, we should let them know why the field isn't showing up as an option with some risk profiles but is with others.)

blanchardjeremy (approved):
Lgtm.

PR #876, opened 2021-06-15
Add vaccine selector for other person

ted107mk:
Adds Their Vaccine selector.

Need clarification on
- Description strings
- Since SegmentedControl's label works differently than SelectControl's, the formatting is inconsistent (whole line is bolded, colon after question mark)

image

### DIFF
--- a/src/data/calculate.ts
+++ b/src/data/calculate.ts
@@ -82,7 +82,7 @@ export const defaultValues: CalculatorData = {
   yourVaccineType: '',
   yourVaccineDoses: 0,
 
-  theirVaccine: 'undefined',
+  theirVaccine: '',
 }
 
 interface CalculatorResult {
--- a/src/data/data.ts
+++ b/src/data/data.ts
@@ -388,14 +388,12 @@ export const Vaccines: { [key: string]: VaccineValue } = {
   },
 }
 
-export const TheirVaccine = {
-  vaccinated: {
-    label: i18n.t('data.their_vaccine.yes'),
-  },
-  unvaccinated: {
-    label: i18n.t('data.their_vaccine.no'),
-  },
-  undefined: {
-    label: i18n.t('data.their_vaccine.unknown'),
-  },
+const yesVaccinatedDesc = i18n.t('data.yes_vaccinated')
+const noVaccinatedDesc = i18n.t('data.no_vaccinated')
+const unknownVaccinatedDesc = i18n.t('data.unknown_vaccinated')
+
+export const TheirVaccine: { [key: string]: CheckBoxFormValue } = {
+  vaccinated: segmentedFormValue(i18n.t('data.their_vaccine.yes'), yesVaccinatedDesc, 1 / 2),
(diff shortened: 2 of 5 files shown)


src/data/calculate.ts
```diff
@@ -82,7 +82,7 @@ export const defaultValues: CalculatorData = {
   yourVaccineType: '',
   yourVaccineDoses: 0,
 
-  theirVaccine: 'undefined',
```
ted107mk, line 85: I was getting a 'Cannot read property 'value' of undefined' js error from SegmentedControl: https://github.com/microcovid/microcovid/blob/main/src/components/calculator/controls/SegmentedControl.tsx#L57

src/data/prepopulated.ts
```diff
@@ -33,7 +33,7 @@ export const prepopulated: {
     theirMask: '',
     yourMask: '',
     voice: '',
-    theirVaccine: 'undefined',
+    theirVaccine: 'undefined', // TODO update this
```
ted107mk, line 36: Will update once default value for theirVaccine is resolved

beshaya (changes requested):
Thanks for doing this! Left some cleanup comments.

I just landed the code for importing vaccine prevalence and using it to compute risk. If you rebase onto main, this should work for the "average person" profile.

beshaya:
Also, we'd love to have you on our slack, which would probably make discussing code easier. Could you fill out https://docs.google.com/forms/d/e/1FAIpQLSeF1r6pHvWyIrZ_quJNzChGOOdjdunVAjrcbqj_xbEUbnMTsA/viewform?usp=sf_link if you're interested?

src/data/data.ts
```diff
@@ -388,14 +388,12 @@ export const Vaccines: { [key: string]: VaccineValue } = {
...
-  },
-  undefined: {
-    label: i18n.t('data.their_vaccine.unknown'),
-  },
+const yesVaccinatedDesc = i18n.t('data.yes_vaccinated')
+const noVaccinatedDesc = i18n.t('data.no_vaccinated')
+const unknownVaccinatedDesc = i18n.t('data.unknown_vaccinated')
+
+export const TheirVaccine: { [key: string]: CheckBoxFormValue } = {
+  vaccinated: segmentedFormValue(i18n.t('data.their_vaccine.yes'), yesVaccinatedDesc, 1 / 2),
```
acconrad, line 396: Is this accurate? I thought other discussions had it all the way down to .2-.16

beshaya: This is a multiplier tedmk found for transmission from vaccinated people who are known to have covid, which is different from the risk reduction from the vaccine.

acconrad: @beshaya but if, in a given area, a small percentage of people have COVID, wouldn't the absolute risk be far lower than 1/2?

beshaya:
This isn't an absolute risk. This is a multiplier much like wearing a mask.

Also, we aren't going to take this particular multiplier, at least until we
see it corroborated by another study or two and think through whether the
other vaccine multipliers need to be adjusted to avoid double counting
reduced contagiousness.

blanchardjeremy:
Closed in favor of #881

PR #887, opened 2021-06-28
Remove theirVaccine if data is unavailable

beshaya:


src/components/calculator/PersonRiskControls.tsx
```diff
@@ -193,10 +202,14 @@ export const PersonRiskControls: React.FunctionComponent<{
...
           useHoverDesc={false}
         />
-      ) : null}
-      <br />
-    </React.Fragment>
-  )
+      )
+    }
+    return (
+      <div>No vaccine data available for this region.</div>
```
blanchardjeremy, line 208: I would say "Since our system doesn't get have vaccine data ifor your region, we aren't able to provide a way for you to adjust your score based on whether or not the person you're seeing is vaccinated."

jy2wong, line 208: translatable string?

jy2wong: "We can't factor in other people's vaccination statuses because we don't have vaccine data for your region."

jy2wong (approved):
Suggestion for how to shorten up the no_vaccine_prevalence string, but I don't insist

blanchardjeremy:
Should we fix the "no data available%" in his PR or a separate one?

blanchardjeremy:
Should we fix the "no data available%" in his PR or a separate one?

6B27142A-B5B0-4687-A294-42B6E1BFB20A

beshaya:
I'd rather do that in a follow-up.

blanchardjeremy (approved):
Didn't do a code review. But did some basic functional testing. Looks good.

PR #888, opened 2021-06-29
Replace node-sass with sass

jy2wong:
Can someone with a Mac make sure this doesn't bork anything?

rm -rf node_modules
yarn install

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
     "markdown-it-footnote": "^3.0.2",
     "markdown-it-github-headings": "^2.0.0",
     "markdown-it-link-attributes": "^3.0.0",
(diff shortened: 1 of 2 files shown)


blanchardjeremy:
When I run Yarn start, I see a whole bunch of sass warnings for a few seconds then they disappear.

Example:
image

jy2wong:
> When I run Yarn start, I see a whole bunch of sass warnings for a few seconds then they disappear.
> 
> Example:
> image

Probably still an improvement IMO, since node-sass seems to be Very Deprecated and (dart) sass seems really far off from 2.0.0.

blanchardjeremy:
Okay I'll follow your lead then! 👍

PR #896, opened 2021-07-06
Change update_prevalence.py to prevent using negative corrections as positive cases.

beshaya:
This may result in lower than expected case counts when a county
reports a correction.


blanchardjeremy (approved):
Looks good to me.

Does this close an existing issue? If so, can you attach it to the PR?

beshaya:
Won't close the issue until we update prevalence.

blanchardjeremy:
Great. I can do it now.

PR #898, opened 2021-07-08
Automatic prevalence update 2021-07-08

blanchardjeremy:


blanchardjeremy:
I can confirm santa clara is working. But it was working before #896 when I did a prevalence update earlier today.

beshaya (approved):

beshaya:
> I can confirm santa clara is working. But it was working before #896 when I did a prevalence update earlier today.

Hah, makes sense. We're now 7 days past the correction that caused the bug.

PR #902, opened 2021-07-12
Gracefully handle missing scenarioName in URLs.

jy2wong:
This commit also cleans up malformed query parameters.

### DIFF
--- a/src/data/calculate.ts
+++ b/src/data/calculate.ts
@@ -16,6 +16,7 @@ import {
   partnerMult,
   personRiskMultiplier,
 } from 'data/data'
+import { PartialData, prepopulated } from 'data/prepopulated'
 
 export interface CalculatorData {
   // Persistence
@@ -124,12 +125,16 @@ export type QueryData = Partial<CalculatorData>
 // Replace any values that no longer exist with empty string (nothing selected).
 // This is used when restoring a previous saved scenario, in case we changed
 // the model in the meantime.
-export const migrateDataToCurrent = (
-  incomingData: Record<string, unknown>,
+// sanitizeData() is used to clean both query parameters in the URL and
+// anything in local storage.
+export const sanitizeData = (
+  data: Partial<CalculatorData>,
+  fillCustomIfScenarioMissing: boolean,
 ): CalculatorData => {
-  const data: Partial<CalculatorData> = { ...incomingData }
   const fixOne = (
-    table: { [key: string]: FormValue | PersonRiskValue | VaccineValue },
+    table: {
+      [key: string]: FormValue | PartialData | PersonRiskValue | VaccineValue
+    },
     prop: keyof CalculatorData,
   ) => {
     const current = data[prop]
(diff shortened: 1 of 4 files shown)


jy2wong:
whoops, found a bug; need to fix before review

blanchardjeremy:
How would you feel about writing a test or two for this?

I'm okay if we publish it without a test. But it would be nice.

src/data/queryParams.ts
```diff
@@ -58,9 +63,13 @@ export const useQueryDataIfPresent = (
...
 ): CalculatorData => {
   const queryDataFiltered = pickBy(queryData, (v) => v !== undefined)
+
   if (isEmpty(queryDataFiltered)) {
     return calcData
   } else {
-    return { ...defaultValues, ...queryDataFiltered }
+    return sanitizeData(
+      queryDataFiltered,
+      true /* fillCustomIfScenarioMissing */,
```
beshaya, line 72: Is there a case where this parameter if `false`? If not, remove parameter?

src/data/calculate.ts
```diff
@@ -124,12 +125,16 @@ export type QueryData = Partial<CalculatorData>
 // Replace any values that no longer exist with empty string (nothing selected).
 // This is used when restoring a previous saved scenario, in case we changed
 // the model in the meantime.
-export const migrateDataToCurrent = (
-  incomingData: Record<string, unknown>,
+// sanitizeData() is used to clean both query parameters in the URL and
+// anything in local storage.
+export const sanitizeData = (
```
beshaya, line 130: I'd love to see a test for these functions. I won't block on it, but it would be really good to have a test file with all the past incarnations of stored data formats.

beshaya: (Also, I low-key prefer these functions be in another file, possibly queryParams. I think the reason for them being structured like this is that queryParams.ts is newer than migrateDataToCurrent. Again, I don't insist.)

beshaya (approved):
LGTM. Would love to see a bit of refactoring + testing, but fine with those being a follow on.

src/data/queryParams.ts
jy2wong: yes.

0. check out code from before the saved scenario UI change
1. run the calculator until you see a points total and something has been saved to local storage
2. hit the reset button to rewrite the thing in local storage with default data
3. note that there is no scenarioName in local storage, so we have the option of filling it with "custom" (allows old saved scenarios and old links to show up properly) or "" (allows reset form to show the typeahead to pick a scenario instead of directly launching you into a custom scenario). I've got it to pick "custom" if the local storage interaction type is nonempty (suggesting that the local storage contains a genuine scenario) or if the data is coming from query params. "" if local storage interaction type is empty.

beshaya: Ah, I see, missed the other call.

src/data/calculate.ts
jy2wong: added tests, maybe refactor later?

blanchardjeremy: Looks like tests are failing. Are you running the tests locally before pushing? If not, let me know if you need any support getting testing working. Took me a bit to get to to speed on it. If so, then I'm wondering why tests pass locally but not on the server.

jy2wong: I did, saw everything pass, and assumed that order of the tests didn't matter (!!!)

Commented out the offending lines, but I'd rather get this in sooner than wait to mock up or clear localStorage to fix it properly.

PR #918, opened 2021-07-29
Use most recent cumulative case count for case count per week

jy2wong:
The offending data that prompted this fix looked like this:

Alameda: daily cumulative cases pulled July 25 2021 with
now = datetime(2021, 7, 25, 12, 0, 0, 0) - timedelta(days=1)

[91028, 91197, 91546, 91743, 91937, 92091, 92372,
 92580, 92584, 92973, 93361, 97459, 93953, 94345]

### DIFF
--- a/update_prevalence.py
+++ b/update_prevalence.py
@@ -260,7 +260,7 @@ class Place(pydantic.BaseModel):
         current = effective_date
         if current not in self.cumulative_cases:
             raise ValueError(f"Missing data for {self.fullname} on {current:%Y-%m-%d}")
-        while len(daily_cumulative_cases) < 14:
+        while len(daily_cumulative_cases) < 15:
             prev = current - timedelta(days=1)
             if prev not in self.cumulative_cases:
                 if prev > min(self.cumulative_cases.keys()):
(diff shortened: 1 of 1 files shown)


blanchardjeremy (approved):
Thumbs up from me. But I only did a quick read and a quick local test.

update_prevalence.py
```diff
@@ -281,16 +281,53 @@ def recent_daily_cumulative_cases(self) -> List[int]:
...
+        negative_corrections = [i for i, val in enumerate(values[:-1]) if val > values[i+1]]
+        corrections_within_bounds = min(values) == values[0] and max(values) == values[-1]
+        if len(negative_corrections) == 0 or corrections_within_bounds:
+            return values[-1] - values[0]
+
+        # Always use values[-1] rather than max(values) because max(values) is
+        # either values[-1] or there's been a negative correction that should take precedence.
+        if min(values[:-1]) <= values[-1]:
+            possibly_suspect_correction = False
+            for local_max_idx in negative_corrections:
```
beshaya, line 295: nit: prefer `negative_correction_index` instead of `local max_idx`, since as you pointed out, we sometimes see negative cases multiple days in a row.

update_prevalence.py
```diff
@@ -281,16 +281,53 @@ def recent_daily_cumulative_cases(self) -> List[int]:
...
+                value_before_correction = values[local_max_idx]
+                value_correction = values[local_max_idx + 1]
+                value_after_correction = None if local_max_idx + 2 >= len(values) else values[local_max_idx + 2]
+
+                if value_after_correction is not None:
+                    correction_size = value_before_correction - value_correction # flipped
+                    overall_change = value_after_correction - value_before_correction
```
beshaya, line 302: Is this basically saying "was the correction reversed out the next day", right? I'd expect "overall change" to mean the difference from day -7 to day 0.

update_prevalence.py
```diff
@@ -281,16 +281,53 @@ def recent_daily_cumulative_cases(self) -> List[int]:
...
+                    # arbitrary heuristic to look for cases where the negative correction
+                    # itself might be the data that's wrong. Warn if:
+                    #   * removing the negative correction would create consistent data,
```
beshaya, line 305: What's meant by "consistent" here?

update_prevalence.py
```diff
@@ -281,16 +281,53 @@ def recent_daily_cumulative_cases(self) -> List[int]:
...
+                    #   * the correction is substantial (more than 5 people) and
+                    #   * the correction is proportionally much larger than the overall change
```
beshaya, line 307: This logic seems a bit odd to me; if the cumulative cases go 5, 10, 0, 13, correction_size is -10, overall change is 3, and this will warn. But I'd look at that as the 0 is just funky.

Meanwhile 5, 10, 0, 14 is fine? Why?

update_prevalence.py
```diff
@@ -894,7 +937,7 @@ def main() -> None:
         data.populate_fips_cache()
 
         # Cumulative cases per region
-        populate_since = effective_date - timedelta(days=15)
+        populate_since = effective_date - timedelta(days=16)
```
beshaya, line 940: Should this still be 15? The code above only looks at index -15 which is 14 elements.

update_prevalence.py
```diff
@@ -281,16 +281,53 @@ def recent_daily_cumulative_cases(self) -> List[int]:
...
-        return max(values[min_index:]) - min(values)
+        # list of indices right before negative corrections. If values = [3,2,3,0,5],
+        # then negative_corrections == [0, 2]
+        negative_corrections = [i for i, val in enumerate(values[:-1]) if val > values[i+1]]
+        corrections_within_bounds = min(values) == values[0] and max(values) == values[-1]
+        if len(negative_corrections) == 0 or corrections_within_bounds:
+            return values[-1] - values[0]
+
+        # Always use values[-1] rather than max(values) because max(values) is
+        # either values[-1] or there's been a negative correction that should take precedence.
```
beshaya, line 292: "precedence" here meaning "warning the person running the script?"

beshaya (approved):
Left some nits but otherwise good to go. This algo probably needs more polish, but I support pushing it out to fix data asap.

update_prevalence.py
jy2wong: monotonically nondecreasing

update_prevalence.py
jy2wong: yeah, it’s not great. But this is very, very noisy without some filtering, and if it’s not off by that much (for negotiable definitions of “that much”, I don’t think manually checking and fixing it will have great impact.

the specific cases I wanted to catch are “oops somebody made a typo and the number is off by thousands”.

1D81184D-359A-4B9A-B519-1231F8F61BED

update_prevalence.py
jy2wong: -15 is 15 elements. You need the 15th day to calculate 7 days worth of cases from cumulative data.

I think it is indeed safe to change this to -15, though.

jy2wong: on Sunday there were 10 cases (cumulative).
on Monday there were 12.
on Tuesday there were 13.
on Wednesday there were 17.

to calculate the number of new cases in 3 days, we need to subtract the Sunday count (index -4) from the Wednesday count (index -1). Otherwise you miss the 2 cases that happened between the end of Sunday and the end of Monday.

PR #923, opened 2021-07-30
Fix typo in formula for suggested budget change after vaccination

hazel-nut:
These values used to be .3 and .03, but with updated data for Delta, they're now .4 and .17 (per the changelog).


beshaya (approved):

beshaya:
Ty for the fix!

src/posts/paper/13-q-and-a.ts
```diff
@@ -230,7 +230,7 @@ As a result, getting a vaccination yourself confers a reduction in incoming micr
 You may additionally increase your budget based on how the vaccine reduces your own chances of negative outcomes, but this is a personal choice dependent on your own view of how averse to the various possible negative outcomes you are.
 
 Some ways you may want to change your budget in after you and your close contacts are vaccinated:
-* If you want to keep the same overall budget but live with housemates, you can increase your personal budget while maintaining the same annual risk; each housemate's activites has a reduced impact on others' microCOVID budget. I.e. if living with N housemates, before vaccination each person's allocation of microCOVIDs would have been (200 / (1 + 0.4N)). After everyone is vaccinated, this increases to (200 / (1 + 0.4*0.17*N)).
+* If you want to keep the same overall budget but live with housemates, you can increase your personal budget while maintaining the same annual risk; each housemate's activites has a reduced impact on others' microCOVID budget. I.e. if living with N housemates, before vaccination each person's allocation of microCOVIDs would have been (200 / (1 + 0.4N)). After everyone is vaccinated, this increases to (200 / (1 + 0.17*N)).
```
beshaya, line 233: I don't think this is right? 

200 / (1 + N * housemate) -> 200 / (1 + N * housemate * vaccine) 

housemate is 0.4, vaccine is 0.17. So it should be 200 / (1 + 0.4 * 0.17 * N)

beshaya (changes requested):

beshaya:
> These values used to be .3 and .03, but with updated data for Delta, they're now .4 and .17 (per the changelog).

They were .3 and .1 * .3 = .03. Now they're .4 and .4 * .17.

src/posts/paper/13-q-and-a.ts
hazel-nut: oh! you're right. the display error on the site is the asterisks getting parsed into italics. better to just put these formulas in code formatting like the ones later on in this section to avoid that. one sec!

blanchardjeremy (approved):
LGTM

PR #927, opened 2021-07-31
Fix preset scenario search.

jy2wong:
It was searching by identifier keys like outdoorMasked2 rather than
"Outdoor masked hangout with 2 other people".


blanchardjeremy (approved):
LGTM

PR #930, opened 2021-08-02
Disambiguate Georgia the state vs country in the prevalence typeahead

jy2wong:


blanchardjeremy (approved):
LGTM. I think we can live without the "nation" situation for now. But happy to have someone integrate

octern:
Agreed re: "nation"; I'd be happy to see this merged and not worry about it further.

PR #932, opened 2021-08-03
Allow "their vaccine" modifier for non-fixed person risk profiles

jy2wong:


src/data/calculate.ts
```diff
@@ -290,29 +291,43 @@ export const calculatePersonRiskEach = (
...
+    } else {
+      // These are risk profiles that were set up for unvaccinated people.
+      switch (data.theirVaccine) {
+        case 'vaccinated':
+          return unadjustedRisk * data.averageFullyVaccinatedMultiplier
+        case 'unvaccinated':
+          return unadjustedRisk
+        // falls through
+        case 'undefined':
+          return unadjustedRisk / data.unvaccinatedPrevalenceRatio
```
beshaya, line 326: I think this should be `* percentVaccinated * averageFullyVaccinatedMultiplier`?

src/data/data.ts
```diff
@@ -322,6 +322,14 @@ export const RiskProfile: { [key: string]: PersonRiskValue } = {
   },
 }
 
+// This is an explicit list of risk profiles for which "their vaccine" modifiers
+// apply.
+export const VaccineModifiableRiskProfiles: (keyof typeof RiskProfile)[] = [
+  'average',
+  'livingAlone',
```
beshaya, line 329: Ostensibly this should work for everything except the fixed value profiles?

src/data/calculate.ts
jy2wong: I basically just followed the existing relationships. If risk profile is average (unadjusted risk means "don't know if they're vaccinated"), then [this line](https://github.com/microcovid/microcovid/pull/932/files/05efc0056efde3345c51d2530aca71ac00225cd3#diff-7d933aaf4cb288c3d417f891c2089c9cd5dbb58ddeb497b5f177b01ae7ea3809L308) says that `unvaccinated = [risk for unknown vax status] * data.unvaccinatedPrevalenceRatio`.

I think the original logic stands and it is correct as written; see [this comment](https://github.com/microcovid/microcovid/pull/932/files/05efc0056efde3345c51d2530aca71ac00225cd3#diff-093196c543f2689ed8ed57c57bb46ce008529f882556657678049ccbc38497f9R360).

src/data/data.ts
jy2wong: that's everything except the fixed microcovid budgets and "literally has covid"?

jy2wong:
This is ready to submit if this comment thread doesn't need any action: https://github.com/microcovid/microcovid/pull/932#discussion_r684724772

src/data/data.ts
beshaya: Yeah, I think so. Thinking aloud here:

 - calculated profiles (not average or fixed) are determined by doing activities with random people and summing the microCOVIDS
 - risk from doing activities = person risk * activity risk
 - vaccine changes activity risk

So I don't see why the vaccine math that applies to livingAlone wouldn't apply to goes to bars?

update_prevalence.py
```diff
@@ -357,14 +357,19 @@ def partial_vaccination_total(self):
...
-    # Average prevalence = sum(prevalence in group * proportion of population in group)
-    #                    = sum(vaccine_mult * unvaccinated_prev * proportion of population)
-    # Unvaccinated risk / Average risk = sum(vaccine_mult * proportion of pop)
-    #                                  = population / sum(vaccine_mult * proportion of pop)
-    # Where the sum is taken over all vaccine types and status (incl no vaccine)
+    # Compute the estimated risk ratio for an unvaccinated person vs an average person,
+    # so that we can convert average risk (which we know from prevalence data) to
+    # unvaccinated risk if we have sufficiently detailed vaccination information.
+    #
+    # risk_sum / population will be the average vaccine multiplier across the
```
beshaya, line 364: I'm having a hard time following this new comment logic.

Could we define risk_sum before using it?

risk_sum = sum(vaccine multiplier * people with that multiplier) + 1 * unvaccinated people
risk_sum / population = avg_vaccine_multiplier
Unvaccinated people have a vaccine multiplier of 1, therefore

unvaccinated_relative_risk = 1 / avg_vaccine_multiplier

update_prevalence.py
```diff
@@ -357,14 +357,19 @@ def partial_vaccination_total(self):
...
+    # entire population, including unvaccinated individuals.
+    # If nobody were vaccinated, the average vaccine multiplier would be 1.
+    # Therefore
+    # unvaccinated_vaccine_multiplier / avg_vaccine_multiplier
```
beshaya, line 368: I find using `unvaccinated_vaccine_multiplier` here confusing because the vaccine multiplier for "no vaccine" is 1.
I was using `unvaccinated_relative_risk`, but we could call it something like `unvaccinated_prevalence_vs_average`.

src/data/calculate.ts
beshaya: Oh, `* (averageVaccine * percentVaccinated + 1 * (1 - percentVaccinated))` drops partially vaxxed, so the formulation you wrote is better.

beshaya (changes requested):

src/data/calculate.ts
beshaya: Could you leave a comment here?

// data.unvaccinatedPrevalenceRatio = population / risk_sum -> 1 / data.unvaccinatedPrevalenceRatio  = average vaccine modifier of the population

src/data/data.ts
```diff
@@ -322,6 +322,12 @@ export const RiskProfile: { [key: string]: PersonRiskValue } = {
   },
 }
 
+// This is an explicit list of risk profiles for which "their vaccine" modifiers
+// apply.
+export const VaccineModifiableRiskProfiles: (keyof typeof RiskProfile)[] = Object.keys(
```
beshaya, line 327: Can we remove this?

beshaya (changes requested):

src/data/data.ts
jy2wong: I can conceive of a future in which we eventually add risk profiles for vaccinated people. Also this line is set in a spot where it picks up everything before the special RiskProfileEnum profiles.

src/data/calculate.ts
jy2wong: I briefly explained what it was and pointed source code readers to the original calculation.

src/data/data.ts
beshaya: > Also this line is set in a spot where it picks up everything before the special RiskProfileEnum profiles.

Oh, yeep, that's super easy to miss.

If the vast majority of profiles are modifiable, I think it would make more sense to have an explicit list of profiles that don't get modified, which would be the keys of RiskProfileEnum below (which is not a very descriptive name).

beshaya (approved):
LGTM % cleaning up VaccineModifiableRiskProfiles.

src/data/data.ts
jy2wong: Renamed RiskProfileEnum to RiskProfilesUnaffectedByVaccines.

jy2wong: (also it would be the values of RiskProfileEnum, rather than the keys)

PR #935, opened 2021-08-04
Add 2% vaccinated budget.

beshaya:
Also remove "dangerously high risk" and use "extremely" instead.

### DIFF
--- a/src/components/calculator/PointsDisplay.tsx
+++ b/src/components/calculator/PointsDisplay.tsx
@@ -131,9 +131,8 @@ export default function PointsDisplay(props: {
     if (name === 'moderate') return t('calculator.category_moderate')
     if (name === 'high') return t('calculator.category_high')
     if (name === 'very-high') return t('calculator.category_very_high')
-    if (name === 'dangerously-high')
-      return t('calculator.category_dangerously_high')
-    if (name === 'extreme') return t('calculator.category_extreme')
+    if (name === 'dangerously-high' || name === 'extreme')
+      return t('calculator.category_extreme')
     return 'Undefined'
   }
   return (
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -109,7 +109,7 @@
     "category_high": "High",
     "category_very_high": "Very High",
     "category_dangerously_high": "Dangerously high",
-    "category_extreme": "Extreme",
+    "category_extreme": "Extremely High",
     "category_postfix": "Risk",
     "prevalence": {
       "reported_prevalence": "Reported prevalence",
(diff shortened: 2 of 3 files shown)


blanchardjeremy:
Also should we change the default budget to 2% (if we're doing that in the blog post?)

src/components/calculator/PointsDisplay.tsx
```diff
@@ -131,9 +131,8 @@ export default function PointsDisplay(props: {
     if (name === 'moderate') return t('calculator.category_moderate')
     if (name === 'high') return t('calculator.category_high')
     if (name === 'very-high') return t('calculator.category_very_high')
-    if (name === 'dangerously-high')
-      return t('calculator.category_dangerously_high')
-    if (name === 'extreme') return t('calculator.category_extreme')
+    if (name === 'dangerously-high' || name === 'extreme')
+      return t('calculator.category_extreme')
```
blanchardjeremy, line 135: I think you maybe have this backward? Dangerously high is below extreme. Here's how they each look: 

image
image

beshaya:
I don't think 2% should be the default budget because "vaccinated" isn't
the default status. We could link them, but that seems like trouble to me.

src/components/calculator/PointsDisplay.tsx
blanchardjeremy: @beshaya so here's a screenshot from the deploy preview with your edits: 
image

I agree that we want to CALL it extreme risk instead of dangerously high.  That makes total sense to me. But I think we want to "demote" extreme from "all red"-mode to just the normal "top of the spectrum.

blanchardjeremy:
@beshaya did we decide we aren't doing this and that it's just a personal choice? or...?

src/locales/en.json
```diff
@@ -169,7 +169,7 @@
     "risk_tolerance_3_percent_label": "Low Caution Budget",
     "risk_tolerance_3_percent_explanation": "3% chance of COVID per year (600 microCOVIDs per week). If you are healthy and prefer to live a riskier lifestyle AND everyone you live with / share air with is willing to do the same.",
     "risk_tolerance_1_percent_label": "Standard Caution Budget",
```
jy2wong, line 171: it might make sense to switch this to "Unvaccinated Caution Budget" or something after 2% appears in the calculator.

jy2wong: also maybe "Unvaccinated Standard Budget" and "Vaccinated Standard Budget"?

src/components/calculator/PointsDisplay.tsx
beshaya: I'm going to remove this from this PR.

src/locales/en.json
beshaya: Done.

jaraco:
I appreciate the change here - guiding users toward selecting a higher risk tolerance based on vaccination. I found myself trying to communicate this factor outside the site and would be grateful to have some accounting for the reasonably increased tolerance for infection due to reduced severity and other treatments afforded by developments.

jaraco (approved):

src/components/calculator/PointsDisplay.tsx
jaraco: This change was apparently removed in one of the force pushes.

src/locales/en.json
```diff
@@ -113,13 +113,32 @@
...
+    "category_dangerously_high": "Dangerously high",
+    "category_extreme": "Extremely High",
+    "category_postfix": "Risk",
     "location_risk_message": "<1>~{{ prevalence }} of people here are currently infectious.</1> We use this number to estimate the baseline chance of transmission per person nearby (given no further information about their behavior).",
     "location_sublabel": {
       "CA": "Province/territory",
       "US": "County",
       "US-AK": "Borough",
       "US-LA": "Parish",
       "country_or_regions": "Region"
```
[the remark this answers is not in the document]
jaraco: I messed up the merge; fixed in 87bd4d2.

PR #936, opened 2021-08-04
Include up-to-date estimate of long COVID prevalence

henryaj:
Update the whitepaper FAQ to include an up-to-date estimate of long COVID prevalence from the ONS and ZOE.


src/posts/paper/13-q-and-a.ts
```diff
@@ -42,9 +42,11 @@ When choosing your budget, you want to factor in  two important aspects: risk to
...
+
+That would mean that a 1% chance of COVID is the same as up to a 0.1% chance of ongoing negative health consequences from COVID. The average American drives [13,476 miles per year](https://www.fhwa.dot.gov/ohim/onh00/bar8.htm). In 2018 [there were](https://cdan.nhtsa.gov/tsftables/National%20Statistics.pdf) 2,491,000 car-crash-related injuries (of any severity) spread across 3,240,327,000,000 vehicle miles travelled (see the [NHTSA website](https://cdan.nhtsa.gov/tsftables/tsfar.htm) for more detailed breakdowns). [A study of Swedish car crash injuries](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3256772/) found that 11.4% of them met the criteria for at least “1% permanent medical impairment” five years later. (“Permanent medical impairment” is a [legally-relevant concept](https://www.disabilitysecrets.com/resources/understanding-permanent-disability-rating-system.htm) that attempts to quantify disability, and the 1% level was the lowest level of disability that was discussed in the Swedish car crash study.) Multiplying this out, we get a 0.11% chance of ongoing negative health consequences from driving for one year.
```
henryaj, line 50: The bulk of the change is here - added a reference to ONS data and the estimates from the ZOE team (which were based on app data after adjusting for the sex and age of the cohort vs the UK population).

Removed reference to "we don't think the ratio is likely to be higher than that" as this is confusing wording. (When scanning I read this as "we don't think it's likely to be higher than 1 in 10", when actually it means the opposite.)

beshaya: The prevailing thinking from the research team right now is that ZOE is underestimating. ZOE is also concerned that people simply stop reporting in the app and they can't control for this. I'm open to including that as another viewpoint if we also include this one: https://nocovid-europe.eu/assets/doc/nocovid_pcs_eng.pdf

> In a recently published study, approximately 30% of patients who had a mild course during acute
COVID-19 illness and were not hospitalized (WHO progression scale 1-3; WHO Working group
2020) experienced persistent or late symptoms consistent with PCS 7 months after infection
(Augustin et al. 2021). 11% of patients were still unable to fully participate in daily life and work 7
months after onset (Augustin et al. 2021).

I agree the "we don't think it's likely to be higher than 1 in 10" is hard to parse. Could you just rewrite it as "A study found that 12% of study participants diagnosed with COVID reported one or more symptoms 24 weeks later" (reading the left side of the chart).

src/posts/paper/13-q-and-a.ts
```diff
@@ -42,9 +42,11 @@ When choosing your budget, you want to factor in  two important aspects: risk to
...
```
beshaya, line 49: Could you also clean up the language here?

That would mean that a 1% chance of COVID is the same as up to a 0.1% chance of ongoing negative health consequences from COVID. -> ...is the same as a 0.12% chance...

beshaya (changes requested):
Thanks for taking care of this. Couple of small comments.

src/posts/paper/13-q-and-a.ts
henryaj: Do you mean the chart on page 8 (figure 2)? I think that's based on the ONS data mentioned earlier.

The ONS data from that study are truncated at 130 days (~18.5 weeks); as you say it seems 12% of people in the COVID-infected group had symptoms at that point. Maybe I drop the ZOE reference and replace the paragraph with the following:

> We think as many as 1 in 10 COVID cases could potentially lead to long-term negative health consequences for the infected person (sometimes called "long COVID"). A study of 20,000 people from the UK's [Office for National Statistics](https://www.ons.gov.uk/peoplepopulationandcommunity/healthandsocialcare/conditionsanddiseases/bulletins/prevalenceofongoingsymptomsfollowingcoronaviruscovid19infectionintheuk/1april2021) showed 13.7% of COVID-infected people continued to experience symptoms after 12 weeks, with 12% continuing to experience symptoms after 18.5 weeks.
>
> That would mean that a 1% chance of COVID is the same as a 0.12% chance ...

henryaj: Have made the change I suggested; let me know your thoughts.

beshaya (approved):

src/posts/paper/13-q-and-a.ts
beshaya: Looks good, thanks for catching that # of weeks.

PR #971, opened 2021-08-16
Move risk budget settings to Step 1

oxytocinlove:
Screen Shot 2021-08-15 at 11 43 44 PM


src/components/calculator/ExplanationCard/CalculationBreakdown.tsx
```diff
@@ -0,0 +1,277 @@
+import React from 'react'
```
oxytocinlove, line 1: Previously, this file was called `ExplanationCard` and contained the calculation breakdown as well as risk budget settings, wrapped in a `Card`. I moved the `Card` to the parent and cleaned this up to just be the `CalculationBreakdown` since we're moving the risk budget settings.

src/components/calculator/controls/RadioControl.tsx
```diff
@@ -23,28 +23,30 @@ export const RadioControl: React.FunctionComponent<{
...
-          <strong>{formValue.label}</strong>
-        </Form.Check.Label>
-        {formValue.sublabel && (
-          <>
-            <Form.Text as="div">{formValue.sublabel}</Form.Text>
-          </>
-        )}
-      </Form.Check>
-    ))}
+    <Accordion activeKey={props.value.toString()}>
```
oxytocinlove, line 26: Turns out `Accordion` Just Works here without needing to have `Card` components

src/components/calculator/prevalence/LocationPrevalenceDetails.tsx
```diff
@@ -31,151 +32,153 @@ export const LocationPrevalenceDetails: React.FunctionComponent<{
...
   return (
-    <ControlledExpandable
-      id={props.id}
-      header={props.header}
-      headerClassName={props.hide ? 'd-none' : ''}
-      open={props.open}
-      setter={props.setter}
-    >
-      <PrevalenceResult data={props.data} />
+    <FormGroup>
```
oxytocinlove, line 35: I just wrapped this in a `FormGroup` to get the spacing right

oxytocinlove:
This PR looks a lot larger than it really is, see comments

beshaya:
Have we had Michael look over this from a user perspective? I think we had
talked about this before and had some concerns that this could overwhelm a
new user? I'm mildly in favor of the change now that we're showing so many
more risk levels.

jy2wong:
I love the accordion-ing of the different radio buttons, but have some further thoughts UX-wise:

1. Conceptually, I think location and risk budget are separate. Information architecture-wise, I think risk budget is really a separate step altogether: step 1 location, step 2 risk budget, step 3 describe the scenario (step "3" here also needs to be broken up IMO, but isn't a top priority for me)

2. We have an explanation of what a risk budget is (up in the starting blurb) that's easy to skip, super far away spatially (this is a problem on mobile), and both too long and not long enough for not-particularly-math-inclined visitors. I am concerned that people will skip the top blurb, see the word budget, and get confused. (Note that this applies to the previous position, too.)

    ...actually, the more I think about it, the more I support a complete reordering. I think a story I'd like to tell is: you're here because you don't know how risky something is. You have a question along the lines of "can I see my family this weekend" or "how risky is going in to work" or "can I dine indoors". I think that's what we should ask first. The next thing to ask would be one of location or budget. With location, an optional explanation that "the riskiness depends on how many people have COVID in your area". With budget, an optional explanation about what a risk budget is and that deciding how big an issue getting infected is is up to you.

This could free us to cut the risk budget Expandable from the top blurb, which I think is a Good Thing. (I am in support of basically anything that shortens the length of the top blurb without adding confusion.)

-----

Would it be possible to separate out the accordion radio buttons change from the position change? I would prefer we get the accordion change in now and do some user studies/apply some design polish before checking in a position change.

PR #976, opened 2021-08-17
Remove prettier/lint for locations.ts

blanchardjeremy:
We're currently getting this error for locations.ts

image

Which is happening because the list/array is supposed to be on multiple lines:

image

I don't know how to resolve this, so I figured I'd just turn off lint. That might be poor form. So I'm open to other ideas.


beshaya:
I'm ok with this, but you could also just run `yarn fix` after `update_prevalence` and it would clean this up (probably. If not this is definitely ok).

blanchardjeremy:
Closing in favor of a yarn-based script approach.

PR #979, opened 2021-08-17
Add prevalence update helper as a yarn script

### LINKED ISSUE #977: Create a yarn script for prevalence updates
Steps would be:
- Use a script similar to #972 to automate the whole process.
- Move the CAN_API_KEY to a local file 
- Update the readme to explain how to set up the `venv` and the API KEY
- Configure a yarn command for prevalence updates
- Add yan fix to the list of commands, since the JSON being generated is currently failing

blanchardjeremy:
Fixes #977


prevalence_helper.sh
```diff
@@ -0,0 +1,34 @@
...
+git branch -D auto-update-prevalence
+git checkout -b auto-update-prevalence
+source .venv/bin/activate
+source .secure-keys
+python3 update_prevalence.py
+
+# Needed in order to fix some lint errors that come up from the JSON generated by update_prevalence.py
+eslint src --fix
+
+# Only proceed if files have have been changed
```
beshaya, line 17: nit: have have -> have

prevalence_helper.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+# TODO(blanchardjeremy): Need to check for the existence of the GitHub CLI before running
```
beshaya, line 3: if ! command -v gh &> /dev/null; then
    echo "GitHub CLI could not be found, please install from https://cli.github.com/"
    exit
fi

KEYSTORE=".secure-keys"
if [ ! -f "$KEYSTORE" ]; then
    echo "$KEYSTORE does not exist. Please obtain an API key from https://apidocs.covidactnow.org/"
    exit
fi

beshaya (approved):
lgtm %checks

beshaya:
I'd vaguely prefer that the script go in a /tools or /scripts directory rather than at the top level. I don't insist.

prevalence_helper.sh
blanchardjeremy: fixed

prevalence_helper.sh
blanchardjeremy: @beshaya how do I then use $KEYSTORE? Do I still do `source .secure-keys` and access `$CAN_API_KEY` or do I access it through the keystore variable somehow?

beshaya:
oh, $KEYSTORE is just a string containing ".secure-keys" to avoid typing it
3 times.

source "$KEYSTORE"

Also, add

set -euo pipefail

to the beginning of the script so that it will stop if any command fails or
an undefined variable is accessed.

blanchardjeremy:
@beshaya Can you  give it another review?

README.md
```diff
@@ -24,6 +24,33 @@ Requirements:
...
+    $ source .venv/bin/activate
+    $ pip install pydantic requests
+    ```
+1. Install the [GitHub CLI](https://cli.github.com/) to be able to using the `gh` command with the `prevalence_helper.sh` script
+    ```sh
+    # using homebrew (see GitHub page for alternate install methods)
+    $ brew install gh
+    ```
+1. Get a [COVIDActNow.org API key from here](https://apidocs.covidactnow.org/). (If you're on the microCOVID team, you can ask for our existing API key.)
+1. Then configure yoru API key as such:
```
beshaya, line 43: yoru -> your

beshaya (approved):
LGTM %nit

PR #1003, opened 2021-08-20
Sort en.json and top level json files by key.

beshaya:


beshaya:
Finding keys in en.json was driving me crazy. Added a plugin to sort them automatically. IIUC, I can't apply this to other language packs without causing lots of trouble for weblate?

blanchardjeremy:
Yeah, I don't think you need to apply it to other language packs because no one is editing those manually. 

This should be fine.

blanchardjeremy (approved):

PR #1006, opened 2021-08-21
Risk profile overhaul

jy2wong:


src/components/calculator/PersonRiskControls.tsx
```diff
@@ -198,32 +173,26 @@ export const PersonRiskControls: React.FunctionComponent<{
...
+      {currentPersonRisk && (
+        <p>
+          <Trans
+            i18nKey="calculator.person_risk_note"
+            values={{ personRisk: fixedPointPrecision(currentPersonRisk) }}
+          >
+            Each nearby person has a person risk of{' '}
+            <strong>200 microCOVIDs</strong> this week. Person risk (chance they
+            currently have COVID) is calculated from their location, behavior,
+            vaccination status, and infection status.
```
blanchardjeremy, line 192: Very simple and clear.

src/locales/en.json
```diff
@@ -97,8 +97,9 @@
     "duration_header": "Duration",
     "duration_question": "How long is the activity, in minutes? (For a repeated activity: minutes per week?)",
     "intimate_risk_warning": "We have applied a minimum Activity Risk for kissing due to the risk involved in exchanging fluids.",
-    "person_risk_profile_question": "What is their risk profile?",
-    "person_risk_profile_header": "Risk Profile",
+    "behavior_question": "How did they spend their last 10 days?",
+    "behavior_header": "Behavior",
+    "person_risk_note": "Each nearby person has a person risk of <2>{{ personRisk }} $t(calculator.pointsdisplay.microCOVIDs)</2> this week. Person risk (chance they currently have COVID) is calculated from their location, behavior, vaccination status, and infection status.",
```
blanchardjeremy, line 102: Is there a way to say this without saying "person risk"? It feels like there might be a way to un-jargon-ify it (since most users aren't going to ever read the white paper or look at the detailed calculation breakdown). 

I wonder about:
* "Each person nearby person has a score of 200 microCOVIDs over the past week" (I know, we also don't use "score" anywhere else in our documentation, so... no ideal)
* Or we could at least put it in quotes or italics. "Each nearby person has a *Person Risk* of 200 microCOVIDs"

blanchardjeremy, line 102: Also, why did you say "this week" vs "currently"? ... When I read "this week" I think "oh, I don' need to go back and re-check this score for a week." When in fact, the prevalence could change tomorrow and their score would change.  I'm sure you were thinking of some edge case to help users understand something, I just don't see it yet. :)

src/locales/en.json
```diff
@@ -277,7 +278,15 @@
       "bars": "Went to a bar in the last 10 days",
       "microcovid_budget_one_percent": "Keeps to a budget of 200 microCOVID per week (\"Standard Caution Budget\")",
       "microcovid_budget_deci_percent": "Keeps to a budget of 20 microCOVID per week (\"High Caution Budget\")",
-      "hasCovid": "Has COVID"
+      "hasCovid": "Has COVID",
+      "notSure_short": "Not sure",
+      "hasCovidRightNow_short": "Infected with COVID",
+      "minimalExposure_short": "1-2 person household, minimal exposure",
+      "minimalExposure": "Once a week groceries; all socializing masked, outdoors and distanced",
+      "worksOrSocializes_short": "Worked or socialized outside home",
```
blanchardjeremy, line 286: Can we get a longer explanation for this one as well?

src/locales/en.json
```diff
@@ -277,7 +278,15 @@
...
+      "largeGathering_short": "Attended large indoor gathering, unmasked",
+      "largeGathering": "Like a conference, night at a bar, or large wedding",
+      "customBudget_short": "Followed a risk budget"
```
blanchardjeremy, line 289: Nit: Slight preference that all of these get organized into an array. (But not important enough to block on it.)

```
"riskProfileTypes": {
    "largeGathering": {
        "title": "Attended large indoor gathering, unmasked",
        "description": "Like a conference, night at a bar, or large wedding" 
    }
    "cusotmBudget": {
        "title": "Attended large indoor gathering, unmasked"
     }
}
```

src/locales/en.json
blanchardjeremy: Upon further review "this week" also seems at odds with the prompt, which is asking about the last 10 days.

src/locales/en.json
```diff
@@ -277,7 +278,15 @@
       "bars": "Went to a bar in the last 10 days",
       "microcovid_budget_one_percent": "Keeps to a budget of 200 microCOVID per week (\"Standard Caution Budget\")",
       "microcovid_budget_deci_percent": "Keeps to a budget of 20 microCOVID per week (\"High Caution Budget\")",
-      "hasCovid": "Has COVID"
+      "hasCovid": "Has COVID",
+      "notSure_short": "Not sure",
+      "hasCovidRightNow_short": "Infected with COVID",
+      "minimalExposure_short": "1-2 person household, minimal exposure",
+      "minimalExposure": "Once a week groceries; all socializing masked, outdoors and distanced",
```
blanchardjeremy, line 285: I mildly want to capitalize AND to emphasize that they are doing all three.

Also, can we get an oxford comma on that list? :)

src/data/data.ts
```diff
@@ -239,25 +240,46 @@ const noContacts = {
...
     ...noContacts,
   },
-  livingAlone: {
-    label: i18n.t('data.person.livingAlone'),
-    personalMultiplier: livingAloneMult,
-    ...noContacts,
-  },
-
   livingWithPartner: {
-    label: i18n.t('data.person.livingWithPartner'),
```
blanchardjeremy, line 255: (This isn't the right spot for this comment, but I'm adding it here because there's no right spot): Can you open an issue to help us keep track of re-doing the pre-made scenarios to use some of these new risk profiles. I think it's odd that "Live-in partner who has no indoor interactions besides you" bumps the risk profile selector to "follows a risk budget" ...  Because that person doesn't follow a risk budget. We just don't have a risk profile that's really appropriate for them. 

Now that I think about it: we could use the 1-2 people living along, minimal risk profile, even though that's technically double-counting some risk points, it would be more intuitive for the user and still get them fairly close to what that number should be. (Seems like we're going for "ballpark" estimations with our risk profiles currently.)

blanchardjeremy (changes requested):
This is great, @jy2wong ! Only one of these comments below is blocking (the longer description for works/socializes). Rest are nits. Thanks so much for making this happen. This is going to help people so damn much.

blanchardjeremy:
@jy2wong and, if you want to get this out the door without any of those changes, I support you, as long as there are issues created to come back to certain things.

src/locales/en.json
xuiqzy: maybe "infection risk of 200 microCOVIDs" or is that not exactly what is meant here?

src/locales/en.json
beshaya: ++

jaraco:
Thanks everyone for the work on this PR. Unfortunately, it appears to have been abandoned with unanswered questions and now has several conflicts. I welcome someone to revive this work if they wish, but for now I'm closing as incomplete.

PR #1019, opened 2021-08-22
Fully automate daily prevalence updates using Github Actions

### LINKED ISSUE #972: Automate the daily prevalence updates using GitHub actions
Basically need to have it function as a cron job so it no longer needs human intervention. 

- Could be done using GitHub actions. 
- Need to store the COVID act now API key as a "secret" for GitHub actions to be able to use it. 
- I think doing it daily makes the most sense.
- Needs error checking to know when to not merge and alert us that something went wrong.

Below is a script I use on my local machine to automate the whole thing. I'm sure the GitHub action configuration will be different. But this might be a useful starting point for whoever codes the GitHub Action.

```

#!/bin/bash

# update local branch
git checkout main
git pull

# delete the remote branch. It usually doesn't exist unless the script failed last time.
git push origin --delete auto-update-prevalence

# delete the local branch
git branch -D auto-update-prevalence

# create the new branch
git checkout -b auto-update-prevalence

# activate the local virtual env.  Need to have previously create it, of course. See update_prevalence.py for set up instructions. 
source .venv/bin/activate

# set the COVIDactnow.org API key (this one is fake). And run the main prevalence script. 
CAN_API_KEY=aaaafffff6666333 python3 update_prevalence.py

# make sure there are actually changes to commit. If the script above failed, there won't be any changes. (probably can do much better error checking than this.)
if [[ `git status --porcelain` ]]; then
  TODAY=`date +%Y-%m-%d`
  git add -A
  git commit -am "Automatic prevalence update $TODAY"
  git pusH
  # use the GitHub CLI to create a pull request and save the url
  PR_URL=`gh pr create --fill`
  echo "Pull request URL: $PR_URL"
  if [[ ! $PR_URL = "" ]]; then
    # set the pull request to auto merge
    gh pr merge --auto --delete-branch --squash "$PR_URL"
  else
    echo "ERROR: No pull request URL generated"
    exit 2
  fi
else
  echo "No changes to add"
fi

```

blanchardjeremy:
- Use GitHub actions to fully automate daily prevalence updates
- Fixes #972


blanchardjeremy:
Research:
- Virtual env: https://github.com/marketplace/actions/restore-or-create-a-python-virtualenv
- Cron job: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events
- Encrypted Secrets (for the API key): https://docs.github.com/en/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-an-environment
- GitHub CLI: https://github.blog/2021-03-11-scripting-with-github-cli/#using-gh-in-github-actions
- Running github actions manually during testing: https://github.community/t/have-a-new-workflow-run-from-a-pr-none-main-branch-before-committing-into-main-branch-possible/195823/6

To do list:
- [ ] Upgrade python version on netlify to 3.x instead of 2.7 (probably by upgrading our build image to Ubuntu Focal)

blanchardjeremy:
Ready for review (thank god!)

blanchardjeremy:
Ugh. Somehow [this commit](https://github.com/microcovid/microcovid/commit/601c8b2072883b0a4cd601bb72d4e32ce51afb8f) messed up the entire formatting of the based on some VSCode automatic formatting I must have turned on. 

I'm not sure to best undo this? Thoughts? I guess it's just a revert of all those changes and re-do the changes I actually intended in that file 


Where are here:

```py
def parse_json(cache: DataCache, model: Type[Model], url: str) -> Model:
    max_attempts = 4
    retry_time_seconds = 60
    for attempt in range(max_attempts + 1):
        # Error case
        if attempt == max_attempts:
            raise ValueError(
                f"Reached max attempts ({attempt}) attmpting to get JSON from {url}"
            )

        # Normal JSON load attmempt
        try:
            contents_as_json = json.loads(cache.get(url))
            break
        except json.JSONDecodeError as e:
            # Lengthen the delay time each time it fails, to give the API more of a break
            # This can lead to very very long script runs, but it (usually) eventually works
            # TODO: We can hopefully greatly redue the delay time once this issue is resolved: https://github.com/andrewthong/covid19tracker-api/issues/88
            current_retry_time_seconds = retry_time_seconds * (attempt + 1)
            print(
                f"JSONDecodeError: {e.msg} at line {e.lineno} col {e.colno}. Document:\n{e.doc}"
            )
            print(
                f"Trying again after {current_retry_time_seconds} seconds ({attempt + 1} attempts so far)..."
            )
            sleep(current_retry_time_seconds)
            cache.remove(url)

    result = pydantic.parse_obj_as(model, contents_as_json)
    return result
```


I've gotta figure out how to disable the formatting on my code editor though....

update_prevalence.py
```diff
@@ -1035,16 +1035,31 @@ def parse_json_list(cache: DataCache, model: Type[Model], url: str) -> List[Mode
...
             contents_as_json = json.loads(cache.get(url))
             break
         except json.JSONDecodeError as e:
-            print(f"JSONDecodeError: {e.msg} at line {e.lineno} col {e.colno}. Document:\n{e.doc}")
-            print(f"Trying again after {retry_time_seconds} seconds ({attempt + 1} attempts so far)...")
-            sleep(retry_time_seconds)
+            # Lengthen the delay time each time it fails, to give the API more of a break
+            # This can lead to very very long script runs, but it (usually) eventually works
+            # TODO: We can hopefully greatly redue the delay time once this issue is resolved: https://github.com/andrewthong/covid19tracker-api/issues/88
+            current_retry_time_seconds = retry_time_seconds * (attempt + 1)
```
beshaya, line 1055: Typically would do exponential backoff, i.e.

retry_time_seconds *= 2

.github/workflows/prevalence.yml
```diff
@@ -1,47 +1,84 @@
...
+
+    - name: Get the current date
+      id: date
+      run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
+
+    - name: Commit changes
+      uses: EndBug/add-and-commit@v7
+      with:
+        branch: auto-update-prevalence
+        message: "Automatic prevalence update ${{ steps.date.outputs.date }}"
```
beshaya, line 72: can we just do "Automatic prevalence update $(date +'%Y-%m-%d')" here, or is that not how github automation works?

.github/workflows/prevalence.yml
```diff
@@ -1,47 +1,84 @@
...
+    - name: Running linter fixes
+      uses: borales/actions-yarn@v2.3.0
+      with:
+        cmd: fix
+
+    - name: Get the current date
+      id: date
+      run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
+
+    - name: Commit changes
```
beshaya, line 68: Is it necessary to use all these modules + steps as opposed to, say

run: yarn update_prevalence -c
env: GITHUB_TOKEN...

.github/workflows/prevalence.yml
```diff
@@ -1,47 +1,84 @@
...
-    
-    # Run update_prevalence.py including the CAN API KEY
-    
-    # Run linter (may need to set up all of yarn first?)
+    # We do this after the update_prevalence script in case we are running it from a branch (during testing)
+    - name: Create a branch
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        git stash
```
beshaya, line 47: Having a hard time following this step. Why are you pushing from local branch main to upstream branch auto-update-prevalence?

.github/workflows/prevalence.yml
blanchardjeremy: It doesn't work like that because this line is a string literal in yaml not an executed line of bash code.

.github/workflows/prevalence.yml
blanchardjeremy: I thought about that at the beginning. Then I went with this so I didn't have to use my own GitHub personal access token. Then I realized the final merge wouldn't work unless I used a different token so at the very end I went with that anyway. So.... Yeah I could try it? I'd have to give it a closer look to make sure there aren't any other unique aspects to the GitHub version.

.github/workflows/prevalence.yml
blanchardjeremy: This is a workaround to help it function when working on a branch (to update . We still wanna start the auto-prevalence-update branch based on main. Then we push to auto-prevalence-update local to remote.

.github/workflows/prevalence.yml
blanchardjeremy: I'll give it a try on a separate branch to see how it goes. Should be easy now that I know how actions work.

Many of the same modules would need to be loaded up though (python, pip, yarn). We wouldn't need the add-and-commit module, I suppose.

update_prevalence.py
blanchardjeremy: Sounds good!

blanchardjeremy:
This is supseded by #1036

blanchardjeremy:
This gist offers a pattern to create a GitHub application (aka: bot) that serves as a bot to edit files and succesfully merge them into your repo, using GitHub actions & workflow files. 

The end result is that you can have workflows that look like this:

image

Instead of this:

image


## Context

This pattern is pretty easy if you are willing to use your own [Personal Access Token (PAT)](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token) to do the commits. But that means your personal user account will show up as the actor for the commits and the pull request merge, which isn't ideal. Also, other workflows on your repo can access your PAT. If you're working on an open-source project, that could pose a security risk to your account.

A better solution is to use an [Installation Access Token](https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps#authenticating-as-an-installation) generated by a GitHub Application. The Application then becomes a "bot" that shows up as the user for all the actions.

If your repository has mandatory checks built-in, the built-in `GITHUB_TOKEN` that most actions use is configured by default to not initiate any further checks (workflows), so as to not create a recursive loop. Which is why we need an alternate approach (see a list of [workarounds](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#workarounds-to-trigger-further-workflow-runs) here).


## Example

If you'd like to see how I used this pattern in my actual situation, you can see the relevant files here:
1. [Cron job workflow](https://github.com/microCOVID/microCOVID/blob/5121f6729d38e20f2aa33f6ed34073cf00327a0c/.github/workflows/prevalence.yml) that initiates the daily file edits
2. [Script that gets called](https://github.com/microCOVID/microCOVID/blob/5121f6729d38e20f2aa33f6ed34073cf00327a0c/scripts/prevalence_helper.sh) from the cron job that actually does the file edits 
3. [Merge workflow](https://github.com/microCOVID/microCOVID/blob/5121f6729d38e20f2aa33f6ed34073cf00327a0c/.github/workflows/prevalence-merge.yml) that merges the pull request. 

(These are permalinks to the files as they were at the time of this writing because I imagine we'll update them moving forward. Feel free to browse our repo to see how we're solving this problem now.)

## This workflow does the following

1. Trigger the workflow from a scheduled cron job (though you can have it run on other events if you wish, just modify the [trigger event](https://docs.github.com/en/actions/reference/events-that-trigger-workflows))
2. Checks out your GitHub repo
3. Makes edits to your files
4. Commits the changes (as the bot)
5. Create a pull request (as the bot)
6. Merges the pull request (as the bot) after all checks are complete

## Installation

1. [Create a GitHub application](https://docs.github.com/en/developers/apps/building-github-apps/creating-a-github-app)
    * You can give it a name. (In this example, it will be named *"Repo Automation"*, but change it to what works for you.)
    * Uncheck **Active** under **Webhook**. You do not need to enter a Webhook URL.
    * Under **Repository permissions: Contents** select **Access: Read & write**.
    * Under **Repository permissions: Pull request** select **Access: Read & write**.
    * (optional) Upload a logo
1. Create a Private key from the App settings page and store it securely.
1. Install your application on the organization/user where your repo lives.
1. Go to **your repository > Secrets > New repository secret** and create `BOT_APP_ID` (with the GitHub Application ID) and `BOT_SECRET_KEY` (with the contents of the secret key you generated)
1. If you have branch protection rules on, go to your repo settings, add the application itself as an authorized committer.
1. Use the two template files below as a basis for what your workflows need to do.
    * You can optionally change the pull request tag from `automerge` to something else. If you do want to modify it, you need to change all instances of that label throughout both files for the script to function. 
1. Merge the files into your default branch for them to become available as options in the **Actions** tab of your repo.
    * After you merge the files in the first, they will become available in the **Actions tab** for you to run manually. Even if you only want to run them on a branch (where you're doing your development), you need to first commit them to your default branch.

Installation notes:

* After you get this script working, you'll be able to look in the git log and discover the `12345678+repo-automation[bot]@users.noreply.github.com` email that GitHub assigned your application. You can then go back and replace the 12345678 with the actual number that should be there. The commits will then show up with the logo that you set.

## Usage

* Option A: Let the cron job run at the specified time
* Option B: Manually run it by going to Repo > Actions > Auto update > Run Workflow

## References

Here are some of the references I found helpful in my quest to get this working.

* [Workarounds to trigger further workflow runs](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#workarounds-to-trigger-further-workflow-runs)
* [Authenticating with GitHub App generated tokens](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens)

PR #1027, opened 2021-08-22
Add option for yarn prevalence to not commit or to stay on current branch

### LINKED ISSUE #1026: Create option for yarn prevalence to not commit changes
Would make running it locally during testing easier

blanchardjeremy:
- Fixes #1026


scripts/prevalence_helper.sh
```diff
@@ -35,24 +62,29 @@ CAN_API_KEY=$CAN_API_KEY python3 update_prevalence.py
...
-  git commit -am "Automatic prevalence update $TODAY"
-  git push
-  # Use the GitHub CLI to create a pull request and save the url
-  PR_URL=`gh pr create --fill`
-  echo "Pull request URL: $PR_URL"
-  if [[ ! $PR_URL = "" ]]; then
-    # Set the pull request to auto merge
-    gh pr merge --auto --delete-branch --squash "$PR_URL"
+
+if [[ $COMMIT_AND_PULL_REQUEST == 1 ]]; then
```
beshaya, line 66: Prefer an early return:

if [[ $COMMIT_AND_PULL_REQUEST != 1 ]]; then
  exit 0
fi

scripts/prevalence_helper.sh
```diff
@@ -10,23 +15,45 @@ if ! command -v gh &> /dev/null; then
...
-git checkout main
-git pull
+if [[ $STAY_ON_CURRENT_BRANCH != 1 ]]; then
+  echo "Creating auto-update-prevalence branch based on main"
+  # Update local branch
+  git checkout main
+  git pull
+
+  # Delete the remote branch. It usually doesn't exist unless the script failed last time.
+  git push origin --delete auto-update-prevalence &> /dev/null || true
```
beshaya, line 50: Curious, what was your reasoning for this to always use the same branch as opposed to, say, "auto-update-prevalence-$(date)"?

blanchardjeremy: Oh just hadn't thought of it. That works!

blanchardjeremy: I still think it might be good to delete the branch, though. In case we're run into problems and need to run the script multiple times. Either that or add hour/minute to the branch name. Thoughts?

blanchardjeremy: i went with hour/minute/secon

PR #1054, opened 2021-08-25
Turn Expandable into an Accordion Card

oxytocinlove:
Call `Expandable`s out a bit more in the UI
Screen Shot 2021-08-25 at 4 54 09 PM


oxytocinlove:
@blanchardjeremy we use `Expandable` in 3 places: The two at the top of the page, the one you noted with the calculation details, and the one that shows the details for your location. IMO it looks good to have them all be this new component, I don't see why we'd want to change some but not all.

oxytocinlove:
@jy2wong I've addressed all the issues you raised. However, in order to make keyboard navigation work, the card header now has a blue outline when it has focus, which IMO doesn't look great. We can remove it, but then it's not clear during keyboard navigation that it has focus, so we need some other way to visually indicate that. Thoughts?

jy2wong:
1. One thing the white background calculation details achieves is clear indication of where the calculation details end (otherwise it's not visually clear that "how to reduce risk" isn't a calculation detail".

    How would it look if the background colour of the expanded content for both prevalence details and calculation details was changed to #F8F9FA (same colour as the old calculation details bg)?

2. I think setting `padding-left: 0.75rem` on the header will help line up the icon with the `[ ] Override location-based data` checkbox.

A reason we'd want to change some but not all is that the surrounding background colour is different between the top-of-calculator and within-calculator expandable. Using the same colours in all places gives the top-of-calculator expandables noticeably more "LOOK AT ME" power.

re: focus: this might help. It's slightly annoying that Safari doesn't support it, but it looks like there's a polyfill?

https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible

blanchardjeremy:
> @blanchardjeremy we use `Expandable` in 3 places: The two at the top of the page, the one you noted with the calculation details, and the one that shows the details for your location. IMO it looks good to have them all be this new component, I don't see why we'd want to change some but not all.

Yeah. I agree that we should use this new format. I think the styling should be different because the grey-on-grey here doesn't create a "box" around the calculation details. Part of the visual "functionality" of this component is that it stands out as distinct from the rest of it's section. Grey-on-grey here doesn't let the user know where this "expanded" box ends.

blanchardjeremy:
Also, just "ditto-ing" the not-great blue outline thing. Especially the corners being cut off.

image

oxytocinlove:
@jy2wong hmm I was not able to get the polyfill working, do you mind giving it a shot

blanchardjeremy:
Looks good. 

For the calculation details: I think you referring it to the "old way" is okay. But another way is to use the same Accordion component and just style this instance of it differently using a class.

blanchardjeremy (approved):
Looks good.

For the calculation details: I think you referring it to the "old way" is okay. But another way is to use the same Accordion component and just style this instance of it differently using a class.

But I won't block on it! So thumbs up to proceed from me.

oxytocinlove:
> For the calculation details: I think you referring it to the "old way" is okay. But another way is to use the same Accordion component and just style this instance of it differently using a class.
> 

I'm confused because that's what I did here (use the same Accordion component and just style this instance of it differently using a class)

blanchardjeremy:
> > For the calculation details: I think you referring it to the "old way" is okay. But another way is to use the same Accordion component and just style this instance of it differently using a class.
> 
> I'm confused because that's what I did here (use the same Accordion component and just style this instance of it differently using a class)

Oh I didn't actually look at the code. I just assumed it was different because the little Chevron styling didn't look as smooth as the other accordions. Gotcha! 

Well good enough to release IMO!

PR #1096, opened 2021-09-04
Add breakdown of activity risk calculation

oxytocinlove:
Screen Shot 2021-09-03 at 11 01 39 PM


src/data/calculate.ts
```diff
@@ -360,25 +362,39 @@ const getVaccineMultiplier = (data: CalculatorData): number | null => {
...
+      return []
     }
-
     const vaccineMultiplier = getVaccineMultiplier(data)
     if (vaccineMultiplier === null) {
-      return null
+      return []
     }
 
+    const forWord = i18n.t('for')
```
oxytocinlove, line 377: Using single words like this is a little hacky, happy to do a more thorough translation job later but am out of energy and wanted to get this up for initial review

src/components/calculator/ExplanationCard/ExplanationCard.tsx
```diff
@@ -42,6 +45,43 @@ const calculationStepHeader = (header: string, value: string): JSX.Element => {
   )
 }
 
+function numberToReadable(value: number): string {
```
oxytocinlove, line 48: This function is pretty hacky but I don't have a better idea - thoughts?

yarn.lock
```diff
@@ -5,14 +5,12 @@
 "@babel/code-frame@7.10.4", "@babel/code-frame@^7.0.0", "@babel/code-frame@^7.10.4", "@babel/code-frame@^7.5.5":
   version "7.10.4"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.10.4.tgz#168da1a36e90da68ae8d49c0f1b48c7c6249213a"
-  integrity sha512-vG6SvB6oYEhvgisZNFRmRCUkLz11c7rp+tbNTynGqc6mS1d5ATd/sGyV6W0KZZnXRKMTzZDRgQT3Ou9jhpAfUg==
```
oxytocinlove, line 8: hmm using `yarn add` seems to have removed all these hashes, any idea why?

src/components/calculator/ExplanationCard/ExplanationCard.tsx
beshaya: It might be a little nicer if num2fraction had a helper function that returned a numerator and denominator rather than splitting here?

src/components/calculator/ExplanationCard/ExplanationCard.tsx
```diff
@@ -42,6 +44,41 @@ const calculationStepHeader = (header: string, value: string): JSX.Element => {
...
+  const fraction = num2fraction(value)
+  const [numerator, denominator] = fraction
+    .split('/')
+    .map((num) => parseInt(num))
+  if (denominator === 1) {
```
beshaya, line 52: Possibly replace this and the next with
```
if (value >= 1) {
  return fixedPointPrecision(value)
}
// get numerator, denominator
if (denominator >= 20) {. // or like, some other threshold
  return fixedPointPercent(value)
}
return fraction.toString()

File: src/data/calculate.ts [hunk 5]
```diff
@@ -394,34 +408,73 @@ export const calculateActivityRisk = (data: CalculatorData): number | null => {
       given: string,
     ): number => (given === '' ? 1 : table[given].multiplier)
 
+    const labelFor = (
+      table: { [key: string]: FormValue },
+      given: string,
+    ): string =>
+      given === '' ? '' : table[given].label_breakdown || table[given].label
```
beshaya, line 415: Since given is a non-nullable string, could this be

!given ? '' : table...

beshaya: Alternatively, could table just include `''` as an entry?

src/data/calculate.ts
```diff
@@ -394,34 +408,73 @@ export const calculateActivityRisk = (data: CalculatorData): number | null => {
...
+  const activityRiskVariables = getActivityRiskVariables(data)
+  if (activityRiskVariables.length === 0) {
     return null
   }
+  let activityRisk = activityRiskVariables
+    .map((variable) => variable.value)
+    .reduce((accumulator, currentValue) => (accumulator *= currentValue))
+
+  if (!['partner', 'repeated'].includes(data.interaction)) {
+    activityRisk = activityRisk / 60.0
```
beshaya, line 468: Why `/ 60.0`? Please comment or use a constant.

src/locales/en.json
```diff
@@ -238,13 +239,18 @@
     "basic_mask": "Multi-layer cloth mask with tight, secure fit",
     "basic_mask_short": "Cloth, thick and snug",
     "car_windows_down": "A moving car with the windows rolled down",
+    "car_windows_down_breakdown": "for car",
```
beshaya, line 242: for airflow from being in a car with the windows down

src/locales/en.json
```diff
@@ -238,13 +239,18 @@
...
     "close_distance": "Close (less than 1 foot / 0.3 meters apart)",
+    "close_distance_breakdown": "for close distance",
```
beshaya, line 244: for being at close distance

src/locales/en.json
```diff
@@ -238,13 +239,18 @@
...
     "filtered": "Indoors with a HEPA filter (flow rate 5x room size per hour)",
+    "filtered_breakdown": "for filtered air",
     "filtered_mask": "Cloth mask with removable filter insert",
     "filtered_mask_short": "Filter Insert",
     "indoor": "Indoor",
     "intimate_distance": "Kissing",
+    "intimate_distance_breakdown": "for kissing",
     "loud_voice": "Loud talking (shouting, talking over music, singing)",
+    "loud_voice_breakdown": "for loud",
```
beshaya, line 253: for increased particles generated by loud talking.

src/locales/en.json
```diff
@@ -257,10 +263,12 @@
     "normal_voice_short": "Normal",
     "oneTime": "A gathering, activity, or errand",
     "outdoor": "Outdoor",
+    "outdoor_breakdown": "for outdoor",
```
beshaya, line 266: for being outdoors

src/locales/en.json
```diff
@@ -257,10 +263,12 @@
...
     "p100_mask_short": "P100 respirator",
     "p100_mask_theirs": "P100 respirator, NIOSH-certified reusable, unclear if outflow valve is covered",
     "p100_mask_yours": "P100 respirator, NIOSH-certified reusable, with covered outflow valve",
     "partially_enclosed": "A space with one or more sides open to the outdoors",
+    "partially_enclosed_breakdown": "for partially enclosed space",
```
beshaya, line 271: for increased airflow from a space with one or more sides open to the outdoors.

src/locales/en.json
```diff
@@ -281,13 +289,24 @@
       "nonFrontline": "An average person in your area who works from home"
     },
     "plane": "An airplane",
+    "plane_breakdown": "for plane",
```
beshaya, line 292: For the airplane's air filtration system

src/locales/en.json
```diff
@@ -281,13 +289,24 @@
...
     "repeated": "Living/staying with others (like roommates or family)",
     "silent_voice": "Not talking (such as quietly riding the train)",
+    "silent_voice_breakdown": "for silent",
```
beshaya, line 295: for reduced aerosol production from not talking

src/locales/en.json
```diff
@@ -281,13 +289,24 @@
...
     "silent_voice_short": "Silent",
     "sixft_distance": "6+ feet / 2+ meters apart",
+    "sixft_distance_breakdown": "for 6+ feet",
```
beshaya, line 298: For being 6+ feet / 2+ meters apart

src/locales/en.json
```diff
@@ -299,6 +318,7 @@
     "thin_mask": "Cloth mask, thin or loose (bandana, buff; or: loose fit, falls down)",
     "thin_mask_short": "Cloth, thin or loose",
     "transit": "A train with air filtration",
+    "transit_breakdown": "for train",
```
beshaya, line 321: for the train's air filtration

yarn.lock
beshaya: Last time I saw this, it was due to running a very old version of yarn.

beshaya (changes requested):
This is a huge improvement, thanks so much.

I left some suggestions for the strings - in general we want to be referencing the text that they saw / selected rather than the internal enum values (for instance the `silent` key is exposed as "not talking" so we should use that language in the explanation). It would also be nice to explain a bit about why the choice is good in the explanation card (for instance, "for the airplanes air filtration" rather than "for plane").

src/locales/en.json
oxytocinlove: that's really long, it'll take up most of the line

src/locales/en.json
oxytocinlove: again really long

src/locales/en.json
oxytocinlove: long

src/locales/en.json
oxytocinlove: long

src/locales/en.json
oxytocinlove: long

src/locales/en.json
oxytocinlove: long

src/data/calculate.ts
oxytocinlove: > This is a huge improvement, thanks so much.
> 
> I left some suggestions for the strings - in general we want to be referencing the text that they saw / selected rather than the internal enum values (for instance the `silent` key is exposed as "not talking" so we should use that language in the explanation). It would also be nice to explain a bit about why the choice is good in the explanation card (for instance, "for the airplanes air filtration" rather than "for plane").

Yeah but that makes them really long which makes the breakdown much harder to read

oxytocinlove:
> This is a huge improvement, thanks so much.
> 
> I left some suggestions for the strings - in general we want to be referencing the text that they saw / selected rather than the internal enum values (for instance the `silent` key is exposed as "not talking" so we should use that language in the explanation). It would also be nice to explain a bit about why the choice is good in the explanation card (for instance, "for the airplanes air filtration" rather than "for plane").

Yeah but that makes them really long which makes the breakdown hard to read

beshaya:
Could we improve that with formatting? Maybe one per line?

blanchardjeremy:
> Could we improve that with formatting? Maybe one per line?


That would be my suggestion as well. Maybe it just needs to be a series of lines rather than trying to put it all on one line. I think it would be easier to read that way even with the short labels. 


So glad his is happening. I think it'll help people who want to understand the underlying model!

jaraco:
Thanks for the contrib here. It looks to have had a good reception, but has outstanding changes requested. This project would be happy to revisit the concerns addressed herein should someone wish to revive the effort. For now, closing as abandoned.

PR #1101, opened 2021-09-05
Add option to jump to error when form data missing

oxytocinlove:
The only required fields in the form are duration and risk profile.

If either is missing, add a "Jump to error" link next to "Fill in calculator to see risk level" that scrolls to and highlights the missing value.

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -138,6 +138,7 @@
     "person_risk_profile_question": "What is their risk profile?",
     "pointsdisplay": {
       "empty_warning": "Fill in calculator to see risk level",
+      "jump_to_error": "Jump to error",
       "microCOVIDs": "microCOVIDs",
       "range": "probably between: {{ from }} to {{ to }}"
     },
--- a/src/components/calculator/DurationInput.tsx
+++ b/src/components/calculator/DurationInput.tsx
@@ -18,6 +18,7 @@ export function DurationInput(props: {
           <Trans>calculator.duration_question</Trans>
         </label>
         <input
+          id="durationInput"
           className="form-control form-control-lg col-md-3"
           type="number"
           inputMode="numeric"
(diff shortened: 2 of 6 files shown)


src/locales/en.json
```diff
@@ -138,6 +138,7 @@
     "person_risk_profile_question": "What is their risk profile?",
     "pointsdisplay": {
       "empty_warning": "Fill in calculator to see risk level",
+      "jump_to_error": "Jump to error",
```
beshaya, line 141: Could we call this "Jump to missing data"

Or "Fill in $FIELD to see results (jump to $FIELD)"

beshaya (approved):
Love this! Left a small wording suggestion, otherwise good to go.

src/locales/en.json
oxytocinlove: Hmm, I considered "Jump to missing data" and actually decided against it, I think it's more confusing (missing data sounds like it's missing on our end)

blanchardjeremy: Could we look to a UI pattern library and see how this is commonly done?

blanchardjeremy: Since @jy2wong often links to Nielsen, I will follow her lead: https://www.nngroup.com/articles/errors-forms-design-guidelines/

Also - https://designmodo.com/ux-form-validation/#4-error-messages

My takeaway from a (very) quick skim of these and a few other articles is basically: attempt to provide inline validation as soon as the error occurs. In my mind, this would mean if they fill out any field after the risk profile field before risk profile has been filled out, highlight it with an error message immediately.

If for some reason we don't go that route,, I wonder about not having any "Jump to error" message and instead just jumping them to it automatically by doing an autoscroll and show the error message then.

oxytocinlove: IMO neither of those options are great - for the first one, they aren't going to be able to see that it's highlighted since they already scrolled past. For the second one - when do we autoscroll? Literally not letting them fill anything else out until they fill out the risk profile seems really annoying

blanchardjeremy: Oh right. There's no "submit" button. I see.... 


Well makes sense how you arrived at this solution, then! 🙃

blanchardjeremy: Personally I'd love it if you could seek out @jy2wong's take on this before publishing. But I don't think I'd block on it.

Issue #721, opened 2021-02-04
Add "how do I log the house all seeing one person" to FAQ


blanchardjeremy:
I thought we had this written, but it doesn't appear to be in there!

vladfi1:
Was about to open an issue for this. The text currently references an "Option B", which doesn't seem to exist.

3geek14:
I too was about to add a similar issue. I'm curious what "Option B" is.

blanchardjeremy:
Oh, we decided to remove that because it was incredibly complicated. I don't remember what we wrote anymore because it was months ago. I will remove the reference to "option B"

blanchardjeremy:
There's a more "budget efficient" way to do it, but it's pretty low on our priority list to add it at this point.

PR #1133, opened 2021-09-19
Clean up 'how to log everyone seeing the same person'

blanchardjeremy:


blanchardjeremy:
Relates to, but doesn't solve #721

PR #1135, opened 2021-09-19
Handle string and float formats for population in Romanian data

oxytocinlove:
The Romanian data seems to have added a bunch of rows where the population is in the format "nnn,nnn.0". The earlier rows are still "nnnnnn" (i.e. parsable as integers).


blanchardjeremy (approved):
LGTM ✅👍

PR #1204, opened 2021-11-07
Add endOfLine: 'auto' to .eslintrc.js

taymonbeal:
Without this change, ESLint outputs a false positive at the end of every line on Windows (even if line endings are otherwise configured correctly).


jaraco:
Boo on eslint for not having sensible defaults.

jaraco (approved):

PR #1205, opened 2021-11-07
Add per-location updated date in preparation for partial updates

oxytocinlove:
Include an updatedAt date in each AppLocation rather than just a single date for all of `location.ts`.

This will allow us to change `update_prevelance.py` to just skip locations that are causing errors rather than failing the entire update.


oxytocinlove:
I tested locally and this works both before and after running `update_prevalence.py`

beshaya (approved):
LGTM

PR #1206, opened 2021-11-07
Pull out each parsing section of update_prevalence.py into its own function

oxytocinlove:
Just a readability improvement in preparation for allowing partial prevalence updates


beshaya (approved):
Nice refactor! Its a pity we don't have a test we could use to make sure no logic changed unexpectedly; maybe run this and the previous version at the same time and sanity check a few locations?

oxytocinlove:
Tested by running `yarn prevalence` once based on main and once based on this branch and seeing that there was no diff

Issue #1226, opened 2021-11-28
B.1.1.529 / Omicron


bluehat:
I didn't see an open thread for the variant yet so here one is. Apologies if this is not how you want to track a developing situation and how it might impact the models. Please let me know and I'll try to do stuff that way.

# Why we might care #
The new variant looks to already have dramatically increased transmission rates. It may also have some degree of immune evasion, though this is yet to be determined.

## Rapid Strain Dominance ##
One clear fact is that this strain has [has dramatically displaced Delta](https://www.nicd.ac.za/wp-content/uploads/2021/11/Update-of-SA-sequencing-data-from-GISAID-26-Nov_Final.pdf) in regions of South Africa. Rates vary seem to be over 70% of new sequenced infections for November, though they are clear that the number of sequenced samples is quite low. 

So, there is probably some reason why this strain is so successful, and that means we probably need to research it. It wouldn't be unreasonable for this to become the new dominant strain in the next few weeks.

[Nature estimates the R in impacted areas of South Africa to be above 2.](https://www.nature.com/articles/d41586-021-03614-z)

# Potential Causes #

## Transmissibility ##
Estimates vary wildly. A bunch of epidemiologists on Twitter are claiming 500% transmissibility over the original strain, compared to Delta's 70%. I have attempted to request these models and will update this if I get them. 

There are a number of weird outbreaks that indicate increased transmissibility.
-  At [a holiday party in Norway](https://www.reuters.com/world/europe/omicron-outbreak-norway-christmas-party-is-biggest-outside-s-africa-authorities-2021-12-03/), half of attendees are confirmed positive. 30 of them were infected after the infection sources left the party.

There are a number of weird outbreaks which should not count as proof of increased transmissibility, only of terrible policy
- On the unfortunate [Netherlands KLM Flight](https://www.bbc.com/news/world-europe-59451103) the passengers were left inside an airplane with no mask enforcement for 5 hours. It is unclear that this situation would have ended much better with previous variants.
- [Another flight to the Netherlands](https://www.francetvinfo.fr/sante/maladie/coronavirus/variant-omicron-le-vol-arrive-a-amsterdam-d-afrique-du-sud-est-une-bombe-a-retardement-s-alarme-un-passager_4861431.html) where people were packed in a waiting room

## Immune Evasion ##
A lot of people are tremendously excited about the number of mutations on this variant and how this might theoretically impact immunity evasion. I have no found any evidence of real-world data to make decisions on at this time.

[...]

PR #1239, opened 2021-12-08
Add warning to indicate that the model has not been updated for Omicron

josiah-roberts:
Copy very much up for revision - I also don't know if we should include a link to additional information in the style of the previous delta notice.

See #1226 and [comment](https://github.com/microCOVID/microCOVID/issues/1226#issuecomment-987135564) regarding the motivation here - it's important that people know what they don't know, and the limitations of the models they are using.


beshaya (approved):

beshaya:
Thanks for getting the ball rolling on this!

PR #1266, opened 2021-12-28
Set up Sentry for both backend and frontend

oxytocinlove:
All unhandled exceptions should now be logged to Sentry.

Also capture printed warnings in `update_prevalence.py`.


beshaya (approved):
LGTM, nice!

PR #1267, opened 2021-12-28
Move location data to a .json file

oxytocinlove:
Data should not be in code files. This also makes it easier to do partial updates and avoids the need for custom formatting when writing the data.


beshaya:
It's pretty hard to review this with all the csv files in the same commit as the code changes. Any chance you could remove the csv files from this pull request or rebase them into a separate commit?

oxytocinlove:
@beshaya done!

oxytocinlove:
(Before merging I'll run the script again to ensure it's not merging in old data)

src/data/locationHelpers.ts
```diff
@@ -1,4 +1,4 @@
-import { Locations, PrevalenceDataDate } from 'data/location'
+import { Locations } from 'data/location'
```
beshaya, line 1: Will importing the json file like this check that types are correct? If not, is there loading we can do to make sure that the data in the file actually matches the PrevalenceData interface?

PR #1279, opened 2022-01-01
Nerfed speaking/shouting to assume only half-time

octern:
Proposed change for the speaking/shouting modifiers. The old formula was modeling someone who was talking nonstop. I changed it to model someone who's speaking / shouting half the time and silent half the time (roughly reasonable if we assume that every person is in a 1:1 conversation and biased high if we assume they're in groups of >2). Discussion at https://microcovid.slack.com/archives/C019MJ8RK45/p1635856803001400

The new multipliers are:
silent: .2x
speaking: .6x
shouting: 4.3x

Note that the "normal" multiplier for the "baseline" person is no longer 1x, which feels a little weird.

### DIFF
--- a/src/data/data.ts
+++ b/src/data/data.ts
@@ -139,12 +139,12 @@ export const Voice: { [key: string]: FormValue } = {
   normal: {
     label: i18n.t('data.normal_voice'),
     label_short: i18n.t('data.normal_voice_short'),
-    multiplier: 1,
+    multiplier: 0.6,
   },
   loud: {
     label: i18n.t('data.loud_voice'),
     label_short: i18n.t('data.loud_voice_short'),
-    multiplier: 5,
+    multiplier: 2.6,
   },
 }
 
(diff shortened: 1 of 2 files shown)


beshaya:
> Note that the "normal" multiplier for the "baseline" person is no longer 1x, which feels a little weird.

ohhh, that reminds me of why these are the way they are. "Normal" comes from contact tracing data assuming doing "normal" things, so it is 1 by definition. Shouting assumes they spend the same portion of time shouting as they would have spent talking. Plausibly this means that "baseline" assumes talking half the time, so the correction would be in the other direction - silent should be 2/5 instead of 1/5 since the baseline already includes not talking half the time.

octern:
Oh, thank you for pointing that out!

Independently, I was also updating the math. Let s be the microCOVIDs
emitted while silent. If we assume that baseline is half talking, then
baseline would be s/2 + 5s/2 = 3s. So the actual silence multiplier would
be 1/3. Does that sound right?

beshaya:
I guess? I think you'd want to make sure we're not making the same mistake
again (i.e. the multiplier already took that into account).

jaraco:
There seems not to be consensus on the proper math here. Given the age of the conversation, I suspect the concern isn't going to be resolved. I'm going to close this PR for now, but feel free to revive the conversation with the same or refreshed proposal.

PR #1282, opened 2022-01-03
Fix noisiest Sentry warnings

oxytocinlove:
Ignore known missing data so that it doesn't warn noisily in Sentry


update_prevalence.py
```diff
@@ -1209,7 +1209,7 @@ def parse_jhu_vaccines_hourly_us(cache, data):
         try:
             state = data.get_state_or_raise(name=item.Province_State, country=item.Country_Region)
         except KeyError:
-            sentry_sdk.capture_message("Could not find state {item.Province_State}")
+            sentry_sdk.capture_message(f"Could not find state {item.Province_State}")
```
oxytocinlove, line 1212: This one was just a typo from my Sentry setup PR, so we weren't getting the names of the states not found. Once fixed I'll debug the actual issues

jaraco:
I'm on the fence whether it's worth merging changes like these or not, given that JHU is no more. Given that we might not see the effect for a long time, if ever, I'm going to decline the change, but we can revisit if we ever revive the data pipelines.

PR #1288, opened 2022-01-05
Handle invalid test positivity rates

oxytocinlove:
Test positivity rate is nullable, so just set it to None if we get an invalid value and log the issue to Sentry instead of blocking the whole update.

### DIFF
--- a/update_prevalence.py
+++ b/update_prevalence.py
@@ -603,7 +603,8 @@ class Place(pydantic.BaseModel):
         if self.test_positivity_rate is not None and (
             self.test_positivity_rate < 0 or self.test_positivity_rate > 1
         ):
-            raise ValueError(f"Positive test rate for {self.name} is {self.test_positivity_rate}")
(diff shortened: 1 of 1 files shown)


oxytocinlove:
This issue is currently blocking the auto-updater. It gets a positivity rate greater than 1 for Quebec.

I spent several hours looking into better solutions. I discovered that:

(1) We are correctly interpreting the Canada Open Covid data, it's just bad for some reason.
(2) There is better data for Quebec and it's even in the same repo maintained by Canada Open Covid, but it's in beta and they explicitly don't trust it.
(3) Pydantic won't let me handle this with a property setter, and also won't let me handle it with a validator.

oxytocinlove:
Right now we are showing a 92.8% positivity rate for Quebec, which is almost certainly false.

justinhaaheim:
This seems reasonable to me. Caveats: I'm new to the repo, and my python is a tad rusty. :)

PR #1293, opened 2022-01-06
Fixing issue 1291 (allow 3 vaccine doses)

RobertSundling:
@beshaya - Here is the fix for issue 1291 the URL query string not allowing 3 vaccine doses


beshaya (approved):

beshaya:
Thanks!

PR #1286, opened 2022-01-04
Omicron vaccine multipliers

amichaelcohn:
Added new multipliers for boosted vaccines + Omicron

PR #1295, opened 2022-01-06
Fix month and typo in changelog

ongardie:
The Date() constructor counts months from 0 to 11, unfortunately.

This changelog entry was added on January 4, 2022 in commit 5c22fbb as part of
PR #1286. Other recent changelog entries appear to have correct months.


beshaya (approved):

beshaya:
Thanks so much for the fix!

PR #1312, opened 2022-01-12
Fix prettier config and format repo

justinhaaheim:
This pr does a few things:

1. Moves the prettier config to .prettierrc.js so that the prettier CLI, vscode and other editors can access it for things like formatOnSave. Previously running `prettier --write .` or `prettier --check .` from the command line would use the _default_ prettier config, ignoring what was in `.eslintrc.js`.
2. Updates the config to apply the same styling across all supported files
3. Upgrades eslint and a few of the lint related packages 
4. Runs prettier on the whole repo (including markdown, json, scss, html) to fix some simple formatting issues


justinhaaheim:
Looks like create react app doesn't like this because the internal eslint version is ^7.5.0. I can downgrade eslint back to the original, or I can upgrade create react app. I'd lean towards the latter, but happy to do either.

justinhaaheim:
I'm still tracking down a few issues surfaced by bumping these libraries. No need to review for now until I've ironed them out.

justinhaaheim:
Say "no" to rabbit holes! I'm going to close this PR and submit a new one with just the prettier fixes with the existing eslint/other libraries. I can put up a separate PR that upgrades the libraries and fixes any issues surfaced.

PR #1313, opened 2022-01-12
Fix risk profiles exceeding one million

### LINKED ISSUE #1310: Certain activities >1 Mn microcovids (e.g., bar = 2 Mn)
**Describe the bug**
For some activities, it seems there isn't a ceiling applies of a max risk factor of 1Mn microcovids.

**Link to microCOVID scenario**
https://www.microcovid.org/?distance=normal&duration=60&interaction=oneTime&personCount=2&riskProfile=average&scenarioName=outdoorMasked2&setting=outdoor&subLocation=US_53033&theirMask=none&theirVaccine=vaccinated&topLocation=US_53&voice=normal&yourMask=none&yourVaccineDoses=3&yourVaccineType=pfizer

**Expected behavior**
Having been at a bar in last 10 days should be capped at 1 Mn, but instead shows as 2 Mn.

**Is this a regression?**
Was this feature working properly before? If so, when was the last time it worked the way you expected?
I'm not sure.

**Additional context / screenshots**
Add any other context about the problem here, including any screenshots.
Screen Shot 2022-01-11 at 10 24 37 AM

justinhaaheim:
This fixes #1310 by limiting the calculation in calculatePersonRiskEach to a maximum of one million.

### DIFF
--- a/.watchmanconfig
+++ b/.watchmanconfig
@@ -1,0 +1,3 @@
+{
+  "root_files": [".watchmanconfig"]
+}
--- a/.watchmanconfig
+++ b/.watchmanconfig
@@ -1,3 +1,0 @@
-{
-  "root_files": [".watchmanconfig"]
-}
--- a/src/data/__tests__/calculate.test.ts
+++ b/src/data/__tests__/calculate.test.ts
@@ -4,6 +4,7 @@ import {
   calculate,
   calculateLocationPersonAverage,
   defaultValues,
+  ONE_MILLION,
 } from 'data/calculate'
 import {
   BUDGET_ONE_PERCENT,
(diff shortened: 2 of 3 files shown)


oxytocinlove (changes requested):
Thanks for taking this on! I think we wanted to use the fix in https://microcovid.slack.com/archives/C019MJ8RK45/p1641845839001200 though

justinhaaheim:
Ah - got it. I'll take a whack at that.

tkpwaeub:
Ok but please note that this isn't really how probability works. If you roll a million sided die a million times the chance of getting a 19 on one of those rolls is roughly 63%. It takes about 4.6 million rolls to clear 99%. I don't see a need for a cap as long as is distinction is explained.

Issue #1317, opened 2022-01-14
Does not support Johnson & Johnson boosters


jrgifford:
The CDC allows mix&match boosters, including a second J&J shot. 

The calculator does not - it only allows you to select one or no shots of J&J. 

![3577EF34-8929-4CD6-84A4-7EF225B18E87](https://user-images.githubusercontent.com/567771/149513137-4668e248-3ce4-4bee-a1b3-57717c902b1c.png)

PR #1318, opened 2022-01-14
Fix J&J multipliers

beshaya:
J&J multiplier was mistakenly listed as .95x for 1 dose with no option for 2 doses. It should have been .95x for 2 doses and 1x for 1 dose.
https://www.medrxiv.org/content/10.1101/2021.12.28.21268436v1.full.pdf

Fixes: #1317


beshaya:
Gonna go ahead and merge this since it's a bugfix.

Issue #1323, opened 2022-01-20
Prevalence update not run since 2022-01-13?


Joegratz:
**Describe the bug**
Until 2022-01-13, prevalence updates were running daily. No prevalence update has happened since then. 

**Link to microCOVID scenario**
https://github.com/microCOVID/microCOVID/pulls?q=label%3Aauto-prevalence-update+

**To Reproduce**
1. Look at last updated date
2. Note that it is more than a week ago

**Expected behavior**
Daily prevalence updates.

**Is this a regression?**
Yes. 

**Additional context / screenshots**
Add any other context about the problem here, including any screenshots.

zestyping:
Hello, I was visiting to report this problem as well.  The last data I see on microcovid.org is from January 12.

![image](https://user-images.githubusercontent.com/236086/150718265-bb6943a9-dff8-4d8e-9cab-652519b86a5e.png)

I made covidmeter.org as an easy way to do faster estimates based on microcovid.org's model.  We fetch data by pulling the `update_prevalence` script from the microCOVID repo and running it, like so:

https://github.com/zestyping/covidmeter/blob/main/get_prevalence_js

It stopped working on January 13 because Antarctica has zero population.  So I simply patched the code to skip Antarctica, and it runs just fine now:

https://github.com/zestyping/covidmeter/commit/611190cd07c50d84b22eeae0d9d30de4ba87009a

You might want to do something similar to get microcovid.org working again.  Thanks for making this project!

justinhaaheim:
This should be fixed now with the PR that was just merged.

beshaya:
@zestyping neat site!

PR #1326, opened 2022-01-23
Prevent missing country data from blocking the prevalence update

justinhaaheim:
Prevalence updates have been failing (#1323) when we try to parse (non-existent?) data for Antarctica. 

It seems reasonable to me in this case to simply log the issue. On the frontend we already [check](https://github.com/microCOVID/microCOVID/blob/main/src/data/calculate.ts#L214-L216) for `population === 0` and return null for prevalence in that case.


justinhaaheim:
So this still fails when we try to call `rollup_cases` for Antarctica later. 

My next thought was that we could wrap the data processing (rollups) for each country in a try catch, and if we catch remove the country from the list (or null it out). Something like that seems useful to keep this script resilient. We'd just want to monitor this fairly closely in case a country starts failing that *should* have data.

But there's still the question of why Antarctica is showing up but with no data in the first place, and perhaps we want to address that head on -- adding a special case to exclude it perhaps?

To me the ideal way that an error like this surfaces and gets handled is that prevalence updates keep going, but we get an alert in slack that a country is failing and then can go investigate it/land a fix, but the prevalence updates haven't been blocked in the process.

update_prevalence.py
```diff
@@ -830,7 +830,7 @@ def rollup_population(parent: Place, child_attr: str) -> None:
                 for child in children.values():
                     parent.population += child.population
                 if not parent.population and parent.name not in fake_names:
```
beshaya, line 832: Could you add Antarctica to fake_names (and perhaps rename fake_names to something like "places_without_population")? Otherwise we'll be spamming ourselves with error logs.

update_prevalence.py
```diff
@@ -830,7 +830,7 @@ def rollup_population(parent: Place, child_attr: str) -> None:
...
-                    print_and_log_to_sentry(f"Missing population data for {parent!r}")
+                    raise ValueError(f"Missing population data for {parent!r}")
```
justinhaaheim, line 833: I think we still need to raise an error here. Why? Because rollup_population is generic and is used for states, counties and countries. If we don't raise an error we might, say, run into a county with no population data and then just leave it be, and then when we're totaling up the state population we'd be missing data for that county and not know it. So I think it's better for the integrity of the data (aka so we can trust the population numbers) to raise an error here and then just catch it further up and handle accordingly.

justinhaaheim:
I think this is ready to go. Here's what I did in this most recent commit:

1. Undo my first commit. See my comment above about why I think we want to still raise an error in that case.
2. Wrapped both rollup passes (aka nested for loops that roll up data) in try catches. If we catch a ValueError, we add that country to a temporary list of `failed_countries` and after we've exited the country for loop we delete those countries from the `self.countries` dict (you can't delete the country from the dict while you're iterating over that dict).

I tested this locally and it the update succeeds, deleting Antarctica because of the "missing population data" error:

```
...

Fetching https://covid19.geo-spatial.org/external/charts_vasile/assets/json/cazuri_zile_long.json... read 27846 objects
⭐️ Discarding country Antarctica due to error: ValueError("Missing population data for Country(fullname='Antarctica', name='Antarctica', population=0, test_positivity_rate=None, cumulative_cases=Counter(), tests_in_past_week=None, vaccines_by_type=None, vaccines_total=Vaccination(partial_vaccinations=0, completed_vaccinations=0), iso3='ATA', states={})")
Warning: Negative correction is suspect; check numbers manually for Unknown, Netherlands. 1 negative cumulative case corrections [3], values=[19717, 20265, 20265, 20977, 19552, 21262, 22012, 23784]
Discarding State(fullname='Dolnoslaskie, Poland', name='Dolnoslaskie', population=2901225, test_positivity_rate=None, cumulative_cases=Counter(), tests_in_past_week=None, vaccines_by_type=None, vaccines_total=Vaccination(partial_vaccinations=0, completed_vaccinations=0), country='Poland', fips=None, counties={}) with no case data

...

$ eslint src --fix && prettier --write src/locales/en.json *.json && black *.py
src/locales/en.json 38ms
package.json 4ms
tsconfig.json 2ms
reformatted update_prevalence.py
All done! ✨ 🍰 ✨
1 file reformatted, 1 file left unchanged.
✨  Done in 14.30s.
```

update_prevalence.py
justinhaaheim: I don't think we want to do that without other changes. That would stop us from throwing for Antarctica, but Antarctica would stay in the list of countries with whatever partial data we calculated. Maybe not a big deal?

What I'd suggest instead is to create a list of `countries_to_ignore` and just purge them in advance of any calculations, so we don't raise any errors in the first place. Or we could rename `fake_names` to `names_to_ignore`, add Antarctica, and then do that same pass through the countries list to remove all the "fake names" before we do any calculations.

For simplicity, and in order to get this merged asap, I'd like to do that in a follow-up PR. Sound good?

update_prevalence.py
```diff
@@ -911,99 +911,127 @@ def rolldown_vaccine_types(parent: Place, children: List[Place]):
...
-                            print_and_log_to_sentry(f"Discarding {county!r} with no case data")
-                        del state.counties[county.name]
-
-                    if county.test_positivity_rate is None:
-                        if county.tests_in_past_week:
-                            county.test_positivity_rate = county.cases_last_week / county.tests_in_past_week
-                            if county.cases_last_week > county.tests_in_past_week:
-                                print_and_log_to_sentry(
-                                    f"Falling back to state data for {county.name} because it has more cases than tests in the last week with a positivity rate of {county.test_positivity_rate}."
+            try:
```
beshaya, line 950: These try-catch blocks suggest the contents could be more readable as a helper function.

update_prevalence.py
```diff
@@ -911,99 +911,127 @@ def rolldown_vaccine_types(parent: Place, children: List[Place]):
...
+                rollup_population(country, "states")
+            except ValueError as err:
+                failed_countries.append(dict(country_name=country.name, error=err))
+
+        # Delete any countries that had an error
+        for country_data in failed_countries:
+            print_and_log_to_sentry(
+                f"Discarding country {country_data['country_name']} due to error: {country_data['error']!r}"
+            )
+            del self.countries[country_data["country_name"]]
```
beshaya, line 944: I'm worried about the user experience if a real country starts having sync problems, would they be confused by suddenly not being able to find their Country? 

Could we do something that keeps the Country name but displays an error about not having data / puts them into manual entry mode / nudges them into manual entry mode?

beshaya (commented):
I like the bold stroke, but I'm worried about people seeing their country go missing... Could we have a UX for that case?

I'm ok with that being a follow-on, but would like to have a plan before merging this.

update_prevalence.py
justinhaaheim: Good point! I have limited time this week, so perhaps we can merge this PR and I'll do this as a follow on?

beshaya (approved):
LGTM with the expectation that we'll improve the UX for removed locations in a follow-up.

justinhaaheim:
> I like the bold stroke, but I'm worried about people seeing their country go missing... Could we have a UX for that case?
> 
> I'm ok with that being a follow-on, but would like to have a plan before merging this.

Yeah, I agree that's not a great UX for a country's data to go missing. *Hopefully* that's fairly rare, judging from the fact that up til now missing country data would have killed the whole prevalence update. I can follow up on this next week.

I should also say that my intention here is not that we _ignore_ missing countries in favor of completing a prevalence update. If it's not already in place I think we should have a way to clearly escalate these missing country logs so that we see when this is happening and can address it directly. That way we're being alerted when data is missing, AND we're still able to continue regular prevalence updates for the rest of the world.

Are we OK to merge this? I don't have merge privileges yet, so @beshaya, @oxytocinlove or someone else will have to merge it.

update_prevalence.py
beshaya: SGTM. 

Another option is to use the population from the last successful update, if applicable.

PR #1328, opened 2022-01-24
Manual prevalence update: 2022 01 24  14 18 24

justinhaaheim:
Here's a manual prevalence update in case we don't want to wait for the next automatic one, given that it's been 11 days since the last one.


justinhaaheim:
Can I get a sanity check?

Data from 1/12/22 for san mateo county:
image

Data from 1/23/22 manual update:
image

The drop in adjusted prevalence (9% to 1.9%) is dramatic! Is that because the adjusted prevalence factors in whether cases are increasing/decreasing? Here's the data from CovidActNow. Cases *were* increasing up until 1/12. Now they've leveled out.

image

https://covidactnow.org/us/california-ca/county/san_mateo_county/?s=28421722

beshaya (approved):

PR #1341, opened 2022-02-02
Configure prettier to run on the whole repo

justinhaaheim:
This PR configures prettier so it runs on all js/ts/json files, not just the ones in `src/locales`, to make the code more uniformly formatted.

To do that we move the prettier config to `.prettierrc.js` so that the prettier CLI, vscode and other editors can access it for things like formatOnSave. Previously running `prettier --write .` or `prettier --check .` from the command line would use the default prettier config, ignoring what was in `.eslintrc.js`.

It also adds a workspace vscode config that contains a launch script to debug the app, and, notably, a settings config that turns on formatOnSave by default. I submit that this is a preferable way to code, since it doesn't require the engineer to fix things manually or run a fix script to handle simple formatting.

If we're a thumbs up on this change I'll add a commit that runs prettier on the whole repo before merging to handle any existing formatting issues.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,10 +9,10 @@ module.exports = {
     react: { version: 'detect' },
   },
   extends: [
-    'plugin:prettier/recommended',
     'prettier/@typescript-eslint',
     'plugin:react/recommended',
     'plugin:@typescript-eslint/recommended',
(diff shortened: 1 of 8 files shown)


jaraco (approved):
This issue has affected me too when I was getting failing checks locally but not in CI.

PR #1391, opened 2022-03-23
remove outdated Omicron warning on calculator

octern:
Omicron is the new normal, and cases are no longer shockingly high, so we should remove the warning.


octern:
Please ignore

Issue #1330, opened 2022-01-27
Add warning to UI when prevalence data is more than ~1 week out of date


justinhaaheim:
When our recent prevalence updates started failing (see discussion on #1326) the data for microcovid.org and the spreadsheet got to be 10 days out of date. Let's create some sort of UI alert to let people know the data is stale and that we're working on fixing it.

apiology:
I was bitten again with the current stall - see https://github.com/microCOVID/microCOVID/issues/1436

I'd be willing to take a crack at this as a first-time contributor.  Let me know what you think or if you have any guidance.

PR #1437, opened 2022-05-14
Warn when prevalence imports are stale

apiology:
Add a warning in calculator that displays when prevalence updates haven't been done in a week or longer.

Fixes https://github.com/microCOVID/microCOVID/issues/1330

Feedback I'd like:

- Does the wording to better match the site's voice?
- Does the timeframe on when to flip on the warning seem about right?

Note: warning only displays once the location is selected, as that's when `calculatordata.prevalanceDataDate` is available.  

A good follow-up PR would be to adjust this to be locale-specific and also handle cases of staleness further upstream.  Note that the 'lastUpdatedDate' entry is [not read in currently](https://github.com/microCOVID/microCOVID/blob/main/update_prevalence.py#L234) - presumably that's the first thing to fix.

image


justinhaaheim:
Thanks for putting up the PR @apiology! I'm supportive of this in general, but I agree it should be closer to the data itself. @beshaya do you have thoughts on this?

apiology:
@justinhaaheim: Sounds good, I'll give that a shot this weekend.

apiology:
Hmmm, looks like at least in some point in the tests, `prevalanceDataDate` becomes a string instead of a Date.  Looking into it.

apiology:
@justinhaaheim @xaqrox @beshaya: I've updated per the feedback above and tests are passing now (GitHub is just waiting for the merge of #1443 for the typecheck-* checks).  

To trigger this while testing manually, I did the following:

1. Go to the bottom of `src/data/location.ts` and change the `updatedAt` attribute of the last entry (for Zimbabwe) to more than a week ago.
2. Load up the site
3. Verify with no location selected no warning appears.
4. Choose a non-Zimbabwe location.
5. Verify no warning appears.
6. Choose Zimbabwe as the location.
7. Verify the warning appears.
8. Click the 'x' on the warning.
9. Verify the warning goes away.

image

src/pages/Calculator.tsx
```diff
@@ -174,26 +174,20 @@ export const Calculator = (): React.ReactElement => {
...
-              <Alert.Heading>
-                {t('calculator.intro.omicron_warning_heading')}
-              </Alert.Heading>
-              <Trans i18nKey="calculator.intro.omicron_warning">
-                Research is in progress to determine Omicron's impact on our
-                models. Preliminary findings suggest an increase in the risk for
-                re-infection.
-              </Trans>
-            </Alert>
-          )}{' '}
```
apiology, line 196: I wasn't sure what the `{' '}` was accomplishing here and did not bring it over.  Let me know if that was a mistake.

src/pages/Calculator.tsx
```diff
@@ -174,26 +174,20 @@ export const Calculator = (): React.ReactElement => {
...
+    prevalenceDataTimeMillis !== null &&
+    currentTimeMillis - prevalenceDataTimeMillis > oneWeekOfMilliseconds
+
+  const showStaleWarning = prevalenceIsStale && !suppressStaleWarning
+
   return (
     <div id="calculator">
       <Row>
         <Col md="12" lg="8" id="calculator-introduction">
-          {showWarning && (
```
apiology, line 181: If you want me to keep this block around for easy drop-in general warnings in the future, just say the word.  Maybe `showGeneralWarning` if so to clarify?  🤷

apiology:
@beshaya Thanks for your help getting #1450 merged - it looks like automatic prevalence updates are moving again 🎊 

If you have the time to review, this one will help warn users for situations that #1450 does not cover.

src/pages/Calculator.tsx
```diff
@@ -174,26 +174,20 @@ export const Calculator = (): React.ReactElement => {
     </button>
   )
 
+  const oneWeekOfMilliseconds = 7 * 24 * 60 * 60 * 1000
```
beshaya, line 177: Could we use moment.js to do math on time instead of converting to milliseconds?

beshaya (changes requested):
Thanks for doing this.

src/pages/Calculator.tsx
apiology: I ended up with [date-fns](https://date-fns.org/); sounds like that's one of the things the [cool kids](https://dockyard.com/blog/2020/02/14/you-probably-don-t-need-moment-js-anymore) are using now that moment.js is [deprecated](https://momentjs.com/docs/#/-project-status/).  date-fns does seem to have nice immutable functions nad play nicely with tree shaking, but I don't have strong feelings about it if you'd prefer moment.js.

I retested using the same method.

beshaya: Looks good to me, thanks!

The hot library of the day moves so fast!

beshaya (approved):

PR #1442, opened 2022-05-15
Remove legacy global PrevalenceDataDate variable

apiology:
I believe it's safe to remove the global PrevalenceDataDate variable from location.ts, which is currently confusing the codebase.  Based on TODOs in the code base, it looks like this was the intent, and based on testing, it appears this was just one last step of a migration that wasn't finished.

Evidence:
* It's not serving a practical purpose; it's not automatically updated by update_prevalence, and so it currently reads December 12, 2021, which is clearly wrong.
* The location-scoped updatedAt in AppLocation in update_prevalance.py is required; the as_app_data() function sets it always.
* Making 'updatedAt' required in the TypeScript code base yields no errors with the current load.


justinhaaheim:
This looks good to me. @beshaya you want to take a look and merge?

beshaya (approved):

PR #1443, opened 2022-05-15
Fix Python type errors, add CI to ensure they stay fixed

apiology:
This addresses existing typechecking failures in the Python code, and adds a (untested so far) GitHub Actions job to validate them going forward.

Note that I found no evidence of real bugs here, but hopefully this will make it easier to stop the next one from sneaking in.


.github/workflows/ci.yml
```diff
@@ -20,13 +20,21 @@ jobs:
...
     - uses: bahmutov/npm-install@v1
     - run: yarn typecheck
+  typecheck-py:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v3
+    - run: pip install -r requirements-manual.txt
+    - run: mypy .
```
apiology, line 37: I'll watch this once you enable the Actions job to run and fix up any problems.

mypy.ini
```diff
@@ -0,0 +1,5 @@
+[mypy]
+warn_unused_ignores = True
+
+[mypy-sentry_sdk.*]
+ignore_missing_imports = True
```
apiology, line 5: Sentry doesn't provide stubs.  A follow-up PR could [add some locally](https://mypy.readthedocs.io/en/stable/stubs.html#stub-files), though.

update_prevalence.py
```diff
@@ -1103,7 +1109,7 @@ def parse_csv(cache: DataCache, model: Type[Model], url: str) -> List[Model]:
 
 def parse_json_list(cache: DataCache, model: Type[Model], url: str) -> List[Model]:
     print(f"Fetching {url}...", end=" ", flush=True)
-    result = pydantic.parse_obj_as(List[model], json.loads(cache.get(url)))
+    result = pydantic.parse_obj_as(List[model], json.loads(cache.get(url))) # type: ignore
```
apiology, line 1112: mypy wants `List[Model]` here, referencing the type instead of the variable as the type parameter; I don't know enough about pydantic to feel confident making that change.

update_prevalence.py
```diff
@@ -1328,11 +1334,11 @@ def get_partially_vaccinated(total_shots, total_fully_vaccinated, shots_for_full
         # add population data
         place = data.get_canada_region_place(region)
         if place.population != 0:
-            raise ValueError(f"Duplicate population info for {place!r}: {region.Population}")
+            raise ValueError(f"Duplicate population info for {place!r}: {region.pop}")
```
apiology, line 1337: This looks like it was wrong beforehand, but would only blow up when we're already handling errors.

update_prevalence.py
```diff
@@ -1399,12 +1405,12 @@ def process_regional_case_reports():
...
                         report.cumulative_testing
                         if min_test_count is None
-                        else min(min_test_count, report.cumulative_testing)
+                        else min(min_test_count, report.cumulative_testing) # type: ignore
                     )
                     max_test_count = (
                         report.cumulative_testing
                         if max_test_count is None
-                        else max(max_test_count, report.cumulative_testing)
+                        else max(max_test_count, report.cumulative_testing) # type: ignore
```
apiology, line 1413: In the above two cases, mypy wasn't smart enough to recognize that in this else statement, the first argument can never be None.

update_prevalence.py
```diff
@@ -1582,21 +1589,21 @@ def format_obj(obj: Dict[str, Any], indent: str = "  ") -> None:
     (csvdir / "date.csv").write_text("Date\n{}".format(effective_date.strftime("%Y-%m-%d")))
     with (csvdir / "index.csv").open("w") as topfile:
         topfile.write("Location,Slug\n")
-        for slug, data in app_locations.items():
-            if not data.topLevelGroup:
+        for slug, location_date in app_locations.items():
+            if not location_date.topLevelGroup:
```
apiology, line 1593: The variable `data` was reused here; mypy doesn't like having the type redefined on an existing variable.

us_state_abbrev.py
```diff
@@ -57,4 +59,9 @@
     "Wyoming": "WY",
 }
 
-us_state_name_by_abbrev = dict(map(reversed, us_state_abbrev_by_name.items()))
+
+def swap(a: Tuple[str, str]) -> Tuple[str, str]:
+    return (a[1], a[0])
+
+
+us_state_name_by_abbrev: Mapping[str, str] = dict(map(swap, us_state_abbrev_by_name.items()))
```
apiology, line 67: mypy didn't seem to be smart enough to understand `reversed` when applied to `Tuple[str, str]`.

requirements-manual.txt
```diff
@@ -1,4 +1,6 @@
 pydantic==1.8.2
 requests==2.25.1
+types_requests==2.27.25
 black==21.8b0
-sentry-sdk==1.5.1
\ No newline at end of file
+sentry-sdk==1.5.11
```
apiology, line 5: sentry-sdk [released a typing fix](https://github.com/getsentry/sentry-python/issues/1415) for an issue run into [in CI](https://github.com/microCOVID/microCOVID/runs/6519153658?check_suite_focus=true)

beshaya (approved):
Wow, this is great, thanks apiology!

apiology:
The missing check above is just because I renamed it to 'typecheck-js' - you may need to adjust project settings to expect 'typecheck-js' and 'typecheck-py' instead of just 'typecheck'.

beshaya:
Thanks for pointing that out. Updated.

apiology:
Ah, it appears the change to use `typing.Counter` triggered https://github.com/samuelcolvin/pydantic/issues/2060 and slipped past my testing.  An upgrade to Pydantic resolved it and `./update_prevalence.py ` runs as expected now.

PR #1444, opened 2022-05-15
DO NOT MERGE - PoC of partial prevalence updates

apiology:
This is (completely manual) proof-of-concept of the idea in https://github.com/microCOVID/microCOVID/issues/1441 - to do a partial update of locations when individual locations fail.

Note that:
* Locations other than the broken Canada feed are updated: 
* A warning message appears when you select a Canadian location:

image

For benefit of myself and other folks that are waiting for prevalence updates, I may try to continue to push prevalence updates up on this branch so we can use the netlify preview site until the updates are unclogged.

PR #1447, opened 2022-05-20
Run manual prevalence update without county-level canada data

justinhaaheim:
Prevalence updates have been failing since 4/30 due to a validation error relating to Canada county-level data. I'm not entirely sure how to fix the validation error, so this prevalence update was run without that data at all -- the idea being that having most of the data up to date is better than having all of the data out of date.


apiology:
Unfortunately, this does take Canada prevalence from lower-than-it-probably-is to zero, which is going to give false assurance.  Of course right now we're giving the whole world false assurance with prevalence numbers that often way lower than they ought to be, so one could argue it's better than the status quo.

I'd suggest using  the manual procedure I laid out in https://github.com/microCOVID/microCOVID/pull/1444#issuecomment-1132795890.  If you want, I can push up a PR using that procedure alone - #1444 combines the procedure with the warning code in #1437

justinhaaheim:
@apiology where are you seeing the canada prevalence data as zero?

It looks to me like the result of commenting out the county-level data is that the microcovid website simply shows country and province as selectable options for canada, and doesn't offer the option to select a county:

image

https://deploy-preview-1447--microcov.netlify.app/

FYI I've pinged @beshaya (one of the main maintainers at this point) to take a look at this and the other PRs. I don't have authorization to merge.

apiology:
@justinhaaheim: Ah!  I was probably making a bad assumption that this was the only data source for Canada--my apologies.  Given that, I should adjust my procedure accordingly in #1444.

PR #1450, opened 2022-05-22
Discard county-level data on validation error for Canada/Romania

### LINKED ISSUE #1441: Skip "county"-level data when their prevalence updates fail
**Is your feature request related to a situation you're trying to understand with microCOVID? Please describe.**

https://github.com/microCOVID/microCOVID/issues/1436 - which caused a lengthy stall in prevalence updates - appears to have been caused by a single location (Canada) having problems during update.

**Describe the solution you'd like**

I'd like to teach `update_prevalence.py` to catch the exception when a location-specific failure happens while fetching county-level data, and still process the other data. 

**Additional context**

This will leave the affected countries with only country and state/province/territory-level prevalence data from ourworldindata.org, but will allow that less-specific data to remain fresh and allow updates for the rest of the world to proceed.

This would affect Canada and Romania, as those are the two countries we fetch county-level prevalence for to layer on top of OWID prevalence data, to my understanding.

I can push up a PR along these lines as time permits, but would welcome input on the approach.

**Relevant Research or Citations**

N/A

apiology:
Fixes #1441

### DIFF
--- a/update_prevalence.py
+++ b/update_prevalence.py
@@ -1304,7 +1304,14 @@ def parse_owid_testing_data(cache, country_by_iso3):
 
 
 def parse_romania_prevalence_data(cache, data):
-    for line in parse_json_list(cache, RomaniaPrevalenceData, RomaniaPrevalenceData.SOURCE):
+    try:
+        romania_data = parse_json_list(cache, RomaniaPrevalenceData, RomaniaPrevalenceData.SOURCE)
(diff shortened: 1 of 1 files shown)


apiology:
@beshaya let me know if this seems like a reasonable approach.  Obv. getting the Canada feed updated to the new format is a good idea regardless; I may play around with that tomorrow if time allows unless someone else is already on top of it.

beshaya (approved):

Issue #1436, opened 2022-05-07
Data hasn't updated in last week


sameerjain0123:
**Describe the bug**
"Data last updated" for San Francisco shows 4/30/2022. My recollection is that data used to update much more frequently, and a lag of a week significantly undercounts risk during a surge like the current one.

**Link to microCOVID scenario**
If applicable, please copy-paste the URL to the scenario that you modeled on the calculator, so we can easily reproduce it.

https://www.microcovid.org/?distance=close&duration=60&interaction=oneTime&personCount=1&riskProfile=livingWithPartner&scenarioName=custom&setting=indoor&subLocation=US_06075&theirMask=thin&theirVaccine=vaccinated&topLocation=US_06&voice=silent&yourMask=n95&yourVaccineDoses=3&yourVaccineType=pfizer

**To Reproduce**
1. Go to microcovid.org
2. Select any country or US State (I tried California / San Francisco, New York, and Hawaii)
3. See error: "Data last updated" is 4/30/2022, even though it is currently 5/7/2022

**Expected behavior**
"Data last updated" should update daily

**Is this a regression?**
Was this feature working properly before? If so, when was the last time it worked the way you expected?

I believe so. I think it worked properly as recently as a week or two ago.

**Additional context / screenshots**
Add any other context about the problem here, including any screenshots.

Refreshing the page, changing / restarting the scenario, toggling variables in the scenario, and trying incognito mode do not resolve the issue.

Screenshot:
![image](https://user-images.githubusercontent.com/90077883/167270823-32e88fbf-2ba5-406b-9108-d6669fe5dc8c.png)

neutrinonerd3333:
I think this is true for all regions (see #1426 and #1427)… but in any case, I don't think there's been any info as to why data updates have stopped recently.

symphonymarie:
In LA at least, the county now only reports bi-weekly unfortunately.

neutrinonerd3333:
Where I am (Santa Clara County, CA), it seems that data is still updated daily, or at least on weekdays. At a glance it seems like the same is true for LA? In any case, I would hope that the data update interval for every region isn't limited by a single region…

apiology:
Running `update_prevalence.py` manually shows that it blows up trying to validate data from Canada - `https://api.opencovid.ca/other?stat=hr`

That URL is 404ing:

```sh
broz@bigbookpro:~/src/microCOVID$ wget https://api.opencovid.ca/other?stat=hr
[...]

PR #1452, opened 2022-05-23
Adjust to breaking changes on https://api.opencovid.ca/

apiology:
This restores health region-level data for Canada and results in `update_prevalence.py` script running successfully to completion.

Fixes https://github.com/microCOVID/microCOVID/issues/1436


update_prevalence.py
```diff
@@ -324,8 +332,6 @@ class CanadaRegionalVaccinationReports(pydantic.BaseModel):
 
     class Report(pydantic.BaseModel):
         date_: date = pydantic.Field(alias="date")
-        total_cases: int
-        total_tests: Optional[int]
```
apiology, line 328: I removed these because they're not used (we only use the vaccination-related items).  Validation started failing because `total_cases` started returning only null.  Perhaps `api.covid19tracker.ca` had the same issue with the breaking API changes from `api.opencovid.ca`.

update_prevalence.py
```diff
@@ -250,18 +250,20 @@ class RomaniaPrevalenceData(pydantic.BaseModel):
...
-class CanadaOpenCovidRegions(pydantic.BaseModel):
-    class RegionInfo(pydantic.BaseModel):
-        HR_UID: int
-        province: str
-        province_full: str
-        health_region: str
-        pop: Union[int, str]
+# Covid Timeline Canada region info dataset:
+class CovidTimelineCanadaRegion(pydantic.BaseModel):
+    SOURCE: ClassVar[str] = "https://raw.githubusercontent.com/ccodwg/CovidTimelineCanada/main/geo/health_regions.csv"
```
apiology, line 255: Health region metadata is no longer hosted via an API, but the data is still around in a CSV we can pull.

update_prevalence.py
```diff
@@ -270,26 +272,28 @@ class RegionInfo(pydantic.BaseModel):
...
 
-class CanadaOpenCovidProvinces(pydantic.BaseModel):
-    SOURCE: ClassVar[str] = "https://api.opencovid.ca/other?stat=prov"
 
-    class Report(pydantic.BaseModel):
-        province: str  # PEI
-        province_full: str  # Prince Edward Island
-        province_short: str  # PE
+class CovidTimelineCanadaProvinceOrTerritory(pydantic.BaseModel):
+    SOURCE: ClassVar[str] = "https://raw.githubusercontent.com/ccodwg/CovidTimelineCanada/main/geo/pt.csv"
```
apiology, line 292: Ditto above - now in CSV form

update_prevalence.py
```diff
@@ -797,8 +803,10 @@ def get_jhu_place(self, jhu_line: JHUCommonFields) -> Place:
         else:
             return self.get_country(jhu_line.Country_Region)
 
-    def get_canada_region_place(self, line: CanadaOpenCovidRegions.RegionInfo) -> Place:
-        return self.get_county(line.health_region, state=line.province_full, country="Canada")
+    def get_canada_region_place(self, health_region: CovidTimelineCanadaRegion, province_or_territory: str) -> Place:
+        return self.get_county(health_region.name_short,
+                               state=province_or_territory,
+                               country="Canada")
```
apiology, line 809: The spelled-out province/territory name isn't provided in this API anymore, but a mapping file is provided, so this method now takes in the full province/territory name.

update_prevalence.py
```diff
@@ -1442,16 +1458,14 @@ def process_regional_case_reports():
...
-                    v
-                    * get_partially_vaccinated(
-                        provincial_reports.summary[-1].cumulative_avaccine,
-                        provincial_reports.summary[-1].cumulative_cvaccine,
-                        2,
-                    ),
-                    v * provincial_reports.summary[-1].cumulative_cvaccine,
+                    manufacturer,
+                    round(proportion * partially_vaccinated),
+                    round(proportion * total_fully_vaccinated),
```
apiology, line 1468: The current API data gives us the number of people who have gotten one, two, three, four, etc shots.

I don't think it'd be correct to pass that information into this function:

```python
    def get_partially_vaccinated(total_shots, total_fully_vaccinated, shots_for_full_vaccination) -> int:
        return total_shots - shots_for_full_vaccination * total_fully_vaccinated
```

The second parameter is meant to be the total number of shots administered in the area.  Perhaps that's what `cumulative_cvaccine` meant (?), but it's certainly not what `vaccine_coverage_dose_2` means.  I believe that `vaccine_coverage_dose_2` means "the number of people who have gotten two doses of the vaccine.  As such, the logic is simpler:

```python
                total_fully_vaccinated = most_recent.vaccine_coverage_dose_2
                partially_vaccinated = most_recent.vaccine_coverage_dose_1 - most_recent.vaccine_coverage_dose_2
```

update_prevalence.py
jeanpaulrsoucy: Note that the `pt_names` and `hr_names` parameters allow the PT and HR names to be returned in a variety of formats for both the `timeseries` and `summary` routes.

update_prevalence.py
jeanpaulrsoucy: Note that both vaccine _coverage_ (% of population) and _administration_ (raw dose numbers) are available. Presently, the former is an adaption PHAC vaccine coverage data (ccodwg/CovidTimelineCanada#21) whereas the latter is an adaption of the PHAC vaccine administration data (ccodwg/CovidTimelineCanada#47). The former `cumulative_cvaccine` value is equivalent to `vaccine_administration_dose_2` in the current dataset.

apiology: @jeanpaulrsoucy: Thanks for the clarification!  Is `cumulative_avaccine` the same as `vaccine_administration_dose_1`?  i.e., the number of people who have gotten at least one shot?

The current microCOVID code assumed that `cumulative_avaccine` counts the total number of shots given, not the number of people who have had at least one shot.

apiology: @jeanpaulrsoucy : One other question - could you clarify the meaning of `vaccine_administration_total_doses`?  It seems to be the sum of the `vaccine_administration_dose_*` values - but if my assumption is correct about what those fields mean, that wouldn't be a number that would have any meaning.

jeanpaulrsoucy: `cumulative_avaccine` is equivalent to `vaccine_administration_total_doses`, which is the sum of all `vaccine_administration_dose_n` values.

update_prevalence.py
```diff
@@ -1352,38 +1368,48 @@ def parse_romania_prevalence_data(cache: DataCache, data: AllData) -> None:
...
+        # their abbreviations
+
+        canada_regions = parse_csv(cache, CovidTimelineCanadaRegion, CovidTimelineCanadaRegion.SOURCE)
+
+        canada_provinces = parse_csv(
+            cache, CovidTimelineCanadaProvinceOrTerritory, CovidTimelineCanadaProvinceOrTerritory.SOURCE
+        )
     except pydantic.error_wrappers.ValidationError as e:
         print_and_log_to_sentry(f"Discarding county-level data from Canada due to error: {e}")
         return
```
apiology, line 1383: I would like to do more here to handle more potential API issues - e.g. right now the only cases that would fail would be the CSVs above going away or columns being renamed.  There are multiple additional API calls below this which are currently unhandled.  I think it would confuse the diff more than is worth it for this PR, however, and there are a couple of different approaches to think through, so I'd propose to punt that for now.

update_prevalence.py
apiology: OK, thanks - I'll dig in on that and double check my assumptions.

apiology: All right, I'm convinced that the logic as I added is correct.  The old logic was correct enough until boosters started to happen, at which point the computed number of partially vaccinated people started trending under the real value, as each booster shot administered inappropriately reduced it by one.

I believe the other (existing) use of `get_partially_vaccinated()` is fine, as it uses api.covid19tracker.ca's `total_vaccinations` and `total_vaccinated` fields, which seem like what the function was intended to use.

apiology:
@justinhaaheim @beshaya: I'm awaiting an upstream data fix at https://github.com/andrewthong/covid19tracker-api/issues/118 before this can be merged, but this is otherwise ready for review.  

Here's what's been done to help make sure this is right:

- Copied in example data from the API responses to the Pydantic definitions to help with review to ensure the right fields were being mapped
- A couple of rounds of self-review of the diff
- Ran `./update_prevalence.py` and verified that Canada health-region-level data appears in `location.ts` and that it appears in the UI
- Verified health region names in `.csv` files seem consistent with past health region names.
- Worked through commentary from @jeanpaulrsoucy, owner of the upstream API in question.
- After discovering an inconsistency in population vs total-people-vaccinated numbers, followed up upstream to find issues there, then verified other numbers in Canada are roughly in line with expectations.

I'd welcome suggestions on other ways to QC this.

Sorry for the delay to all who were waiting to get this functionality back!  It's been an interesting learning process getting familiar with microCOVID's data ingestion.

apiology:
Update: https://github.com/andrewthong/covid19tracker-api/issues/118 is resolved!  This is ready to merge once approved.

update_prevalence.py
beshaya: This seems right to me; previously cumulative_avaccine was all we could get from the API.

beshaya (approved):

beshaya:
Thanks for the ping! And thanks for the fix!

Issue #1476, opened 2022-06-19
Prevalence is stale


apiology:
The prevalence warning popped up today:

<img width="357" alt="image" src="https://user-images.githubusercontent.com/3681194/174490002-3334e692-9f8d-4eed-b929-1941c60d5dfa.png">

This appears to be the last update PR that was opened: https://github.com/microCOVID/microCOVID/pull/1475

apiology:
@beshaya: Heads-up on this one.  It's possible that I'll be able to dig in on this one this weekend, but I have not had time yet.

apiology:
```sh
$ ./update_prevalence.py 
Fetching https://raw.githubusercontent.com/CSSEGISandData/COVID-19/master/csse_covid_19_data/UID_ISO_FIPS_LookUp_Table.csv... read 4318 objects
Fetching https://raw.githubusercontent.com/CSSEGISandData/COVID-19/master/csse_covid_19_data/csse_covid_19_daily_reports/06-24-2022.csv... read 4012 objects
[...]
Discarding State(fullname='Presov, Slovakia', name='Presov', population=798596, test_positivity_rate=None, cumulative_cases=Counter(), tests_in_past_week=None, vaccines_by_type=None, vaccines_total=Vaccination(partial_vaccinations=0, completed_vaccinations=0), country='Slovakia', fips=None, counties={}) with no case data
Discarding State(fullname='Trencin, Slovakia', name='Trencin', population=600386, test_positivity_rate=None, cumulative_cases=Counter(), tests_in_past_week=None, vaccines_by_type=None, vaccines_total=Vaccination(partial_vaccinations=0, completed_vaccinations=0), country='Slovakia', fips=None, counties={}) with no case data
Discarding State(fullname='Trnava, Slovakia', name='Trnava', population=554172, test_positivity_rate=None, cumulative_cases=Counter(), tests_in_past_week=None, vaccines_by_type=None, vaccines_total=Vaccination(partial_vaccinations=0, completed_vaccinations=0), country='Slovakia', fips=None, counties={}) with no case data
Discarding State(fullname='Zilina, Slovakia', name='Zilina', population=694763, test_positivity_rate=None, cumulative_cases=Counter(), tests_in_past_week=None, vaccines_by_type=None, vaccines_total=Vaccination(partial_vaccinations=0, completed_vaccinations=0), country='Slovakia', fips=None, counties={}) with no case data
Positive test rate for New Zealand is 1.5345431391276663
[...]

PR #1478, opened 2022-06-25
Drop county-level data from Romania if stale

apiology:
Adds a check to validate whether Romania county-level source is up to date, and discard Romania county-level data if not.

Fixes https://github.com/microCOVID/microCOVID/issues/1476


apiology:
@beshaya, would you be so kind as to review?

update_prevalence.py
```diff
@@ -1362,6 +1362,11 @@ def parse_romania_prevalence_data(cache: DataCache, data: AllData) -> None:
     except pydantic.error_wrappers.ValidationError as e:
         print_and_log_to_sentry(f"Discarding county-level data from Romania due to error: {e}")
         return
+    latest_date = max([region.Date for region in romania_regions])
+    if effective_date > latest_date:
```
beshaya, line 1366: Should we give a day or two of leeway here?

beshaya (commented):
Thanks so much for looking into this! Left one question.

update_prevalence.py
apiology: @beshaya: Unfortunately, it's not as easy as adding a day or two to that `if` statement--the script is built with the assumption that we can always join data against the latest date from JHU.   Later on in processing, we blow up if the effective_date is not found -  https://github.com/microCOVID/microCOVID/blob/main/update_prevalence.py#L419-L421 - which is what was stopping updates.
 
Note that this problem doesn't bother us with api.opencovid.ca, as we pass the "fill=true" flag, causing it to just duplicate the most recent day's actual data as needed up to the current date.  Filling is not a fantastic solution when it's out of our control--the site could go indefinitely stale and we'd have no way to know.  Potentially that approach could be used on our side for both countries, and ideally in a way that preserves the real `updatedAt` date, but I'd prefer to focus on getting the updates running again in this PR.

beshaya: Ok, fair enough. Thanks for thinking this through. I agree that getting updates running again is top priority.

beshaya (approved):

Issue #1570, opened 2022-09-19
Prevalence updates failing


apiology:
**Describe the bug**

Automatic prevalence updates have started failing - the first failure started on  the `auto-update-prevalence-2022-09-17--14-21-18` branch.

**Link to microCOVID scenario**

https://github.com/microCOVID/microCOVID/actions

**To Reproduce**

1. Set up per https://github.com/microCOVID/microCOVID/#local-development
2. Run `./update_prevalence.py`
3. See error:
 
```sh
Traceback (most recent call last):
  File "./update_prevalence.py", line 1685, in <module>
    main()
  File "./update_prevalence.py", line 1608, in main
    app_locations[state.app_key] = state.as_app_data()
  File "./update_prevalence.py", line 692, in as_app_data
    result = super().as_app_data()
  File "./update_prevalence.py", line 627, in as_app_data
    last_week = self.cases_last_week
  File "./update_prevalence.py", line 494, in cases_last_week
    return self.cases_in_cum_cases(self.recent_daily_cumulative_cases[-8:])
  File "./update_prevalence.py", line 419, in recent_daily_cumulative_cases
    raise ValueError(
ValueError: Missing data for Pitcairn Islands, United Kingdom on 2022-09-18 - Counter({datetime.date(2022, 9, 13): 4, datetime.date(2022, 9, 12): 4, datetime.date(2022, 9, 11): 4, datetime.date(2022, 9, 10): 4, datetime.date(2022, 9, 9): 4, datetime.date(2022, 9, 8): 4, datetime.date(2022, 9, 7): 4, datetime.date(2022, 9, 6): 4, datetime.date(2022, 9, 5): 4, datetime.date(2022, 9, 4): 4, datetime.date(2022, 9, 3): 4, datetime.date(2022, 9, 2): 4})
```

**Expected behavior**

Prevalence updates successfully.

**Is this a regression?**

Yes; last success was 2022-09-14

**Additional context / screenshots**

<img width="865" alt="image" src="https://user-images.githubusercontent.com/3681194/191012833-cdf64b01-5103-45a7-9a2b-c71cff3daacb.png">

apiology:
Perhaps related to this upstream issue https://github.com/CSSEGISandData/COVID-19/issues/6096

Fixed upstream in https://github.com/CSSEGISandData/COVID-19/commit/789b82db2b9e6a33ac64be8e137f1d8b68646046 - prevalence just ran successfully and the site is now up to date.

PR #1571, opened 2022-09-21
Cleanly drop state data upon failures and continue processing

apiology:
Fixes https://github.com/microCOVID/microCOVID/issues/1570

This will drop a state-level locale from the dataset if we encounter an error while processing data within it.  #1570 was caused by the upstream JHU data no longer sending data from Pitcairn Islands, which is considered a state-level locale under the UK in the JHU dataset.

### DIFF
--- a/update_prevalence.py
+++ b/update_prevalence.py
@@ -1604,10 +1604,27 @@ def main() -> None:
     for country in sorted(data.countries.values(), key=namegetter):
         app_locations[country.app_key] = country.as_app_data()
         for state in country.states.values():
-            if state.app_key not in app_locations:
-                app_locations[state.app_key] = state.as_app_data()
-            for county in state.counties.values():
-                app_locations[county.app_key] = county.as_app_data()
+            try:
+                state_data: Optional[AppLocation] = None
+                county_data: Dict[str, AppLocation] = {}
+                if state.app_key not in app_locations:
+                    state_data = state.as_app_data()
+                for county in state.counties.values():
+                    county_data[county.app_key] = county.as_app_data()
+                # once everything succeeded, go ahead and attach
+                if state_data is not None:
+                    app_locations[state.app_key] = state_data
+                for key, appdata in county_data.items():
+                    app_locations[key] = appdata
+            except ValueError as e:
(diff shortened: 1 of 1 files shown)


update_prevalence.py
```diff
@@ -1604,10 +1604,27 @@ def main() -> None:
...
+                for key, appdata in county_data.items():
+                    app_locations[key] = appdata
+            except ValueError as e:
+                print_and_log_to_sentry(f"Couldn't calculate {state}: {e}")
+
+                # calculator expects all subdivisions to have data populated
+                if state.app_key in app_locations[country.app_key].subdivisions:
+                    print_and_log_to_sentry(
+                        f"Removing {state} from app_locations[{country.app_key}].subdivisions"
+                    )
```
apiology, line 1626: The original issue that motivated this was https://github.com/microCOVID/microCOVID/issues/1570, which only affected the Pitcairn Islands - population 47 people - and left the site stale for a week until JHU noticed and resolved the issue.

If we had a working alert system, we could use the population affected to adjust error handling.  We could define something akin to DEBUG/INFO/WARNING/ERROR thresholds, where data issues affecting populations above a certain size go to WARNING (actively alerting us) or even ERROR (where it would fail the prevalence run like it does today as well as alerting us).

Failing prevalence would represent a choice to leave the site with stale data rather than significantly limited data.

update_prevalence.py
```diff
@@ -1604,10 +1604,27 @@ def main() -> None:
     for country in sorted(data.countries.values(), key=namegetter):
         app_locations[country.app_key] = country.as_app_data()
         for state in country.states.values():
-            if state.app_key not in app_locations:
-                app_locations[state.app_key] = state.as_app_data()
-            for county in state.counties.values():
-                app_locations[county.app_key] = county.as_app_data()
+            try:
```
coachnate, line 1607: Ha! @apiology - now I see the try/catch 👍

apiology: Hi @coachnate!  This is a different PR than your original comment in https://github.com/microCOVID/microCOVID/pull/1655.  This approach would not have had any effect on https://github.com/microCOVID/microCOVID/issues/1654.

Issue #1575, opened 2022-09-24
Non-US vaccination data is missing


apiology:
**Describe the bug**

Non-US countries have lost "Percent of population fully vaccinated" data.

**Link to microCOVID scenario**

* [Example scenario from UK](https://www.microcovid.org/?distance=sixFt&duration=120&interaction=oneTime&personCount=8&riskProfile=average&scenarioName=1person_15minCarRide&setting=indoor&theirMask=none&topLocation=United_Kingdom&voice=normal&yourMask=n95&yourVaccineDoses=3&yourVaccineType=pfizer)

**To Reproduce**
1. Load microCOVID calculator and enter a non-US entry in the country field.
2. Unfold 'Details'
3. Look at 'Percent of population fully vaccinated' field
4. Observe that 'no data available' appears

**Expected behavior**
Numeric percentage appears

**Is this a regression?**
I believe it started happening a few days ago; will add a comment with what I've found digging in on this so far.

**Additional context / screenshots**

Add any other context about the problem here, including any screenshots.

<img width="303" alt="image" src="https://user-images.githubusercontent.com/3681194/192099691-9f71d41e-2831-41fc-8328-941b67f5097b.png">

apiology:
One of [microCOVID's data sources](https://github.com/microCOVID/microCOVID/blob/main/update_prevalence.py#L190), the JHU global vaccine data CSV, [was archived on 2022-09-21]([https://github.com/govex/COVID-19/tree/master/data_tables/vaccine_data/global_data/archive]), probably related to the 'U.S. & Global Vaccine Data Change' section of https://github.com/CSSEGISandData/COVID-19/issues/6082

Similar data appears [here](https://github.com/govex/COVID-19/tree/master/data_tables/vaccine_data/global_data) - perhaps it could be used.

Related:
* Vaccination has a very limited effect on risk of spread regardless in our model, so this is probably not a big effect on results for folks.
* I only noticed this while poking around https://github.com/CSSEGISandData/COVID-19/issues/6082 and https://github.com/CSSEGISandData/COVID-19/issues/6099 to understand the impact to microCOVID after seeing a news article about it.
* `update_prevalence.py` silently ignores 404s from fetched data and treats the result as an empty data file, which appears to be what happened here.
[...]

PR #1578, opened 2022-09-25
Adapt to JHU vaccine data changes

apiology:
Fixes https://github.com/microCOVID/microCOVID/issues/1575


update_prevalence.py
```diff
@@ -1269,25 +1301,79 @@ def parse_jhu_daily_report(cache: DataCache, data: AllData, current: date) -> No
...
+    # People with a First Booster Dose: 109,578,270
+    # People with a Second Booster Dose: 23,118,101
+    #
+    # Let's assume that these trends can be applied world-wide,
+    # and thus if n people had at least one dose,
+    # 224980931/263812108 = 85% of them completed the primary
+    # series.
+    primary_series_completion_ratio = 0.85
+
+    return round(partial_vaccinations * primary_series_completion_ratio)
```
apiology, line 1329: Since JHU is no longer aggregating better information from its sources, and the current microCOVID model uses (or at least reports on) primary series completion, it's up to us to come up with an estimate of it based on what they give us now.

I tested this way of estimating against an initial try where I came up with a very conservative possible estimate of vax rate (see that initial code here: https://github.com/microCOVID/microCOVID/pull/1578/files#diff-093196c543f2689ed8ed57c57bb46ce008529f882556657678049ccbc38497f9R1332-R1344)

France vax rate before: 81%
France vax rate after - conservative: 59%
France vax rate after - CDC-adjusted: 71% 
Notes: I haven't dug up ground truth to see what number is more correct here, but we're at least in the ballpark of the numbers JHU was giving out before.

US vax rate before: 68%
US vax rate after - conservative: 47%
US vax rate after - CDC-adjusted: 69%
Notes: CDC says 67.8% of US population completed primary series: https://covid.cdc.gov/covid-data-tracker/#vaccinations_vacc-people-additional-dose-totalpop - so the number before and the CDC-adjsuted numbers are as expected, given this was the training set for this super rough method.

Alabama vax rate before: 52%
Alabama vax rate after - conservative: 52%
Alabama vax rate after - CDC-adjusted: 52%
Notes: We continue to get person-level data for primary series completion for US states through JHU, not just at-least-one-dose-received, so this is as expected.

update_prevalence.py
```diff
@@ -1122,6 +1149,11 @@ def parse_csv(cache: DataCache, model: Type[Model], url: str) -> List[Model]:
     reader = csv.reader(lines)
     fields = [re.sub("^7", "Seven", re.sub("[^A-Za-z0-9_]", "_", name)) for name in next(reader)]
 
+    # verify we have every field we expect
+    for field_name in model.__fields__:
+        if field_name not in fields:
+            raise ValueError(f"Did not find expected '{field_name}' column in header of {url}")
```
apiology, line 1155: If a column which is marked Optional in our model, the source could slightly rename the column and we'd silently start treating the entire column as None.  This block makes us more sensitive to those sorts of structural changes.

PR #1580, opened 2022-09-25
Improve CSV-parsing typesafety by not assuming zero values

apiology:
update_prevalence.py's current CSV parsing code assumes that a CSV data element with no value (think the second entry in `foo,,bar,1,baz`) maps to a 0 if we defined field as a non-optional float or int.

I'd prefer we treat these as validation errors, which forces us to use Optional[int]/Optional[float], which mypy will force us to handle the None case when the value is returned.

I hope that this will alert us earlier if there are assumptions about upstream data that don't hold for some data or have been violated due to upstream changes.


apiology:
Tested by diffing `src/data/location.ts` and `public/prevalence_data/*.csv` between before and after runs with this change.  There were no differences.

update_prevalence.py
```diff
@@ -1239,6 +1244,11 @@ def parse_jhu_place_facts(cache: DataCache, data: AllData, country_by_iso3: Dict
             # has its own entry so turn the combo into just Bristol Bay
             line.Admin2 = "Yakutat"
             line.Population = 604  # from Google
+        if line.Population is None:
+            if line.Admin2 == "Unassigned":
+                line.Population = 0
+            else:
+                raise ValueError(f"Please manually provide population for {line}")
```
apiology, line 1251: Example case - mypy forces me to add this `if` statement.  The special case additions were found experimentally by running the script until it completed successfully.

update_prevalence.py
```diff
@@ -1205,6 +1205,8 @@ def ignore_jhu_place(line: JHUCommonFields) -> bool:
         "Micronesia",
         "Palau",
         "Summer Olympics 2020",
+        "Winter Olympics 2022",
```
apiology, line 1208: Note that there are a handful of things like this that may pop up over time, which could result in a failed run until we update the blocklist here.  I'd prefer to get a better failed-run notification system in place before merging this PR to mitigate this risk.  If along the way we can get a concept of a 'warning' that becomes equally visible, but discards the locale without failing the run, perhaps there would be room to make this a warning instead.

Issue #1654, opened 2022-11-26
Prevalence runs failing


apiology:
Our daily prevalence updates [have failed the last four days (11/23-11/26)](https://github.com/microCOVID/microCOVID/actions).  This is an example log, which is unfortunately hidden to folks who aren't part of the org:

```
13s
Run yarn prevalence -b -c -k *** -v /home/runner/.virtualenvs/.venv
yarn run v[1](https://github.com/microCOVID/microCOVID/actions/runs/3532758468/jobs/5927512982#step:10:1).22.19
$ ./scripts/prevalence_helper.sh -b -c -k *** -v /home/runner/.virtualenvs/.venv
Using a manual virtualenv directory: /home/runner/.virtualenvs/.venv
Branch will be based on the currently checked out branch
Switched to a new branch 'auto-update-prevalence-2022-11-23--14-0[7](https://github.com/microCOVID/microCOVID/actions/runs/3532758468/jobs/5927512982#step:10:8)-30'
remote: 
remote: Create a pull request for 'auto-update-prevalence-2022-11-23--14-07-30' on GitHub by visiting:        
remote:      https://github.com/microCOVID/microCOVID/pull/new/auto-update-prevalence-2022-11-23--14-07-30        
remote: 
To https://github.com/microCOVID/microCOVID
 * [new branch]          auto-update-prevalence-2022-11-23--14-07-30 -> auto-update-prevalence-2022-11-23--14-07-30
branch 'auto-update-prevalence-2022-11-23--14-07-30' set up to track 'origin/auto-update-prevalence-2022-11-23--14-07-30'.
Created branch auto-update-prevalence-2022-11-23--14-07-30
Activating the virtualenv
Activating virtulenv: /home/runner/.virtualenvs/.venv/bin/activate
Running prevalence script
Fetching https://raw.githubusercontent.com/CSSEGISandData/COVID-19/master/csse_covid_19_data/UID_ISO_FIPS_LookUp_Table.csv...
read 4322 objects
Fetching https://raw.githubusercontent.com/CSSEGISandData/COVID-19/master/csse_covid_19_data/csse_covid_19_daily_reports/11-22-2022.csv...
read 4017 objects
Fetching https://raw.githubusercontent.com/CSSEGISandData/COVID-19/master/csse_covid_19_data/csse_covid_19_daily_reports/11-21-2022.csv...
read 4017 objects
[...]

PR #1655, opened 2022-11-27
Log to sentry and continue if JHU vaccine data is unavailable

shawnbiesan2:
This PR allows the prevalance script to continue running in the case of missing JHU Vaccination Data by logging to Sentry instead of throwing an exception


shawnbiesan2:
This is the very literal "keep prevalance running" fix. Not sure if there should be more thought put into how the overall model is impacted when JHU Vaccine data is unavailable.

coachnate:
I just dropped a comment on the issue related to this. I was thinking about adding some try/catch blocks in there. Thoughts?

shawnbiesan2:
Closing as https://github.com/microCOVID/microCOVID/issues/1654 was resolved by upstream

Issue #1084, opened 2021-08-30
Remove label description for environment


jy2wong:
Environment: and the dropdown items themselves should suffice, and user testing suggests the indoor/outdoor questions may increase confusion.

Al-0:
I will work on this!

mariehuynh:
To clarify, is this asking for "Is it indoor or outdoor? Is the space ventilated, or is the air filtered?" to be removed?

apiology:
That's my reading of it.

mariehuynh:
Would you rather have that field simply disabled or removed completely from all languages?

mariehuynh:
Ah, I just noticed README-TRANSLATIONS.md. Would you rather me leave the field there but blank in case of future changes?

apiology:
To clarify, I believe this is asking for the somewhat redundant description to be removed, not the actual dropdown itself.  

As far as the nuts and bolts, I'd say propose what you think makes sense, maybe describe your reasoning, and I can dig in at the PR review process - not sure I've read README-TRANSLATIONS.md myself :)

PR #1661, opened 2022-12-02
issue #1084

mariehuynh:
Simplest change that leaves the field available for future text insertions. Translations will propagate automatically via Weblate


apiology:
Thanks, @mariehuynh!

PR #1680, opened 2022-12-06
Log when no cases found for Place in update_prevalence.py

apiology:
Log for analysis Places where no new cases were noticed at all for a
week.  These may be places with spotty or no current reporting that
we should not take at face value.


apiology:
This takes our current WARNING summaries to:

```
WARNING: 19,237,682 people affected by No county-level case data
WARNING: 38,246,108 people affected by No tests last week
WARNING: 16,840,340 people affected by Endpoints say no new cases and max(values) says new cases
WARNING: 67,427,787 people affected by No state-level case data
WARNING: 271,562,497 people affected by No cases noted for a week - State level
WARNING: 909,848,811 people affected by No cases noted for a week - Country level
WARNING: 48,101,095 people affected by No cases noted for a week - County level
WARNING: 4,822,233 people affected by Invalid test positivity rate
WARNING: 661,636 people affected by Negative correction is suspect
```

apiology:
I think something's quirky with Python's branch coverage reporting, as the missing coverage it's reporting back is non-sensical 🤷

PR #1689, opened 2022-12-10
Configure Sentry with updated key

shawnbiesan2:


apiology (approved):
Thanks!!

setup.cfg
```diff
@@ -7,5 +7,9 @@ omit =
      node_modules/*
      .venv/*
 
+[coverage:report]
+exclude_lines = 
+     def configure_logging
+
```
shawnbiesan2, line 13: I don't see a great way to truly test configure_logging so seems reasonable to exclude it from coverage calculation

Issue #1509, opened 2022-07-19
Data for Canada health regions in British Columbia are all zeros


justinhaaheim:
**Describe the bug**
![image](https://user-images.githubusercontent.com/3389743/179651502-adf7946d-f1ed-4d15-ba72-a9aed1edb99c.png)

0 cases reported in the last week for Canada > BC > Interior

**Link to microCOVID scenario**

https://www.microcovid.org/?distance=sixFt&duration=120&interaction=oneTime&personCount=15&riskProfile=average&scenarioName=bar&setting=indoor&subLocation=Canada_British_Columbia&subSubLocation=Canada_British_Columbia_Interior&theirMask=none&topLocation=Canada&voice=loud&yourMask=none

apiology:
Hey @jeanpaulrsoucy, any idea what might be going on here?  Here's an example where `cases_daily` returns 0:

```sh
$ curl 'https://api.opencovid.ca/summary?loc=BC&ymd=true&before=2022-07-18&after=2022-07-02' | jq . | tail -40
############################################################################ 100.0%
      "vaccine_administration_dose_1": 4524033,
      "vaccine_administration_dose_1_daily": 0,
      "vaccine_administration_dose_2": 4357594,
      "vaccine_administration_dose_2_daily": 0,
      "vaccine_administration_dose_3": 2446320,
      "vaccine_administration_dose_3_daily": 0
    },
    {
      "region": "BC",
      "date": "2022-07-18",
      "cases": 375938,
      "cases_daily": 0,
      "deaths": 3823,
      "deaths_daily": 0,
      "hospitalizations": 426,
      "hospitalizations_daily": 0,
      "icu": 34,
      "icu_daily": 0,
      "tests_completed": 6110205,
      "tests_completed_daily": 0,
      "vaccine_coverage_dose_1": 86.75,
      "vaccine_coverage_dose_1_daily": 0,
      "vaccine_coverage_dose_2": 83.56,
      "vaccine_coverage_dose_2_daily": 0,
      "vaccine_coverage_dose_3": 46.91,
      "vaccine_coverage_dose_3_daily": 0,
      "vaccine_coverage_dose_4": 5.18,
      "vaccine_coverage_dose_4_daily": 0,
      "vaccine_administration_total_doses": 11598028,
      "vaccine_administration_total_doses_daily": 0,
      "vaccine_administration_dose_1": 4524033,
      "vaccine_administration_dose_1_daily": 0,
[...]

PR #1691, opened 2022-12-12
Update api.opencovid.ca call for HR case data

jeanpaulrsoucy:
This PR is my attempt to address #1509.

Since transitioning to the `CovidTimelineCanada` dataset, the date structure of the dataset changed (see [Transitioning from `Covid19Canada`](https://github.com/ccodwg/CovidTimelineCanada/blob/main/docs/transitioning_from_Covid19Canada.md) for more details). Different PTs/HRs now have different end dates, based on their update schedules.

In this PR, I have made the following changes:

- The call for HR case data now returns the most recent 19 days of data available for that HR, and the "before" and "after" parameters have been removed.
- I did not touch the call for the PT summary data, since data in the `summary` route always ends with the current date, unlike the `timeseries` route.

I got 19 days by inspecting the provenience of the "before" and "after" variables:

https://github.com/microCOVID/microCOVID/blob/f1bbc621aca41126de2084cd14724840b72da7d0/update_prevalence.py#L1621-L1623

The "effective date" appears to be calculated as two days ago at most,

https://github.com/microCOVID/microCOVID/blob/f1bbc621aca41126de2084cd14724840b72da7d0/update_prevalence.py#L142-L147

...so 16 days before the effective date should be 19 days ago at most.

I am not familiar with the rest of the codebase, so I'm not sure if having different date ranges for different HR data (i.e., the most recent 7 days of data from Alberta and Manitoba will not be the same) will cause issues or not. If it's all index-based, it shouldn't cause issues.

### DIFF
--- a/update_prevalence.py
+++ b/update_prevalence.py
@@ -392,7 +392,7 @@ class CovidTimelineCanadaRegion(pydantic.BaseModel):
 class CanadaOpenCovidCases(pydantic.BaseModel):
     SOURCE: ClassVar[
         str
-    ] = "https://api.opencovid.ca/timeseries?stat=cases&loc={hr_uid}&geo=hr&ymd=true&fill=true&before={before}&after={after}"
+    ] = "https://api.opencovid.ca/timeseries?stat=cases&loc={hr_uid}&geo=hr&date=19"
 
     class Data(pydantic.BaseModel):
         class Report(pydantic.BaseModel):
@@ -1685,9 +1685,7 @@ def parse_canada_prevalence_data(cache: DataCache, data: AllData) -> None:
                 cache,
                 CanadaOpenCovidCases,
                 CanadaOpenCovidCases.SOURCE.format(
-                    hr_uid=region.hruid,
-                    before=canada_effective_date.strftime("%Y-%m-%d"),
-                    after=populate_since.strftime("%Y-%m-%d"),
+                    hr_uid=region.hruid
                 ),
             )
             for report in case_reports.data.cases:


jeanpaulrsoucy:
Checking the deploy preview, we still see 0s for British Columbia. However, based on the commit history, it seems prevalence data updates happen separately, so the updated script for Canadian data hasn't actually been run. Will need @apiology to test.

apiology:
@jeanpaulrsoucy: You can run `update_prevalence.py` locally and push up the location.ts once updated - there are instructions here: https://github.com/microCOVID/microCOVID/#prevalence-updates

You can ignore all the .csv files if you like - those are used for the spreadsheet.

Happy to help if you run into trouble - maybe we find some time to pair on it?  https://calendly.com/apiology/microcovid

jeanpaulrsoucy:
I followed the instructions in the README and managed to run `yarn prevalence -b` on my new branch. Unfortunately, I got an error:

> Traceback (most recent call last):
>   File "/home/jprs/Desktop/microCOVID/update_prevalence.py", line 1922, in <module>
>     main()
>   File "/home/jprs/Desktop/microCOVID/update_prevalence.py", line 1845, in main
>     app_locations[county.app_key] = county.as_app_data()
>   File "/home/jprs/Desktop/microCOVID/update_prevalence.py", line 738, in as_app_data
>     last_week = self.cases_last_week
>   File "/home/jprs/Desktop/microCOVID/update_prevalence.py", line 605, in cases_last_week
>     return self.cases_in_cum_cases(self.recent_daily_cumulative_cases[-8:])
>   File "/home/jprs/Desktop/microCOVID/update_prevalence.py", line 529, in recent_daily_cumulative_cases
>     raise ValueError(
> ValueError: Missing data for South, Alberta, Canada on 2022-12-11 - Counter({datetime.date(2022, 12, 5): 40485, datetime.date(2022, 12, 4): 40474, datetime.date(2022, 12, 3): 40472, datetime.date(2022, 12, 2): 40464, datetime.date(2022, 12, 1): 40454, datetime.date(2022, 11, 30): 40437, datetime.date(2022, 11, 29): 40430, datetime.date(2022, 11, 28): 40414, datetime.date(2022, 11, 27): 40402, datetime.date(2022, 11, 26): 40384, datetime.date(2022, 11, 25): 40381, datetime.date(2022, 11, 24): 40367, datetime.date(2022, 11, 23): 40358, datetime.date(2022, 11, 22): 40344, datetime.date(2022, 11, 21): 40339, datetime.date(2022, 11, 20): 40326, datetime.date(2022, 11, 19): 40320, datetime.date(2022, 11, 18): 40313, datetime.date(2022, 11, 17): 40302})
> Sentry is attempting to send 2 pending error messages
> Waiting up to 2 seconds
> Press Ctrl-C to quit
> error Command failed with exit code 1.

The run fails for first health region, Alberta - South. From the error log, it's easy to see why: it's still expecting data from certain fixed dates, rather than just the last x days available.

The API query for the relevant HR data:

```
https://api.opencovid.ca/timeseries?stat=cases&geo=hr&loc=4831&date=19
```

...returns data from 2022-11-17 to 2022-12-05, the most recent date of data available. It looks like the script expects data as far back as 2022-11-17 (which we have covered), but expects data up to 2022-12-11 (i.e., the "effective date"), which we obviously don't.

I'm wondering if scrubbing the effective date for Canada altogether and purely using indexing will fix the issue, but I'm not sure how much work this would require.

apiology:
@jeanpaulrsoucy: That's right - currently update_prevalence.py bases its data structures on the dates pulled from JHU.  I don't think that's strictly required to calculate microCOVIDs, but it is how update_prevalence.py works right now.

That said, I think there's a larger change here to address #1509

Our [model](https://www.microcovid.org/paper/7-basic-method) wants to know how many new cases a place reported, calculated on a weekly basis.

It does this by subtracting the most recent number of cumulative positive tests reported from that reported a week ago - https://github.com/microCOVID/microCOVID/blob/main/update_prevalence.py#L603-L605

It does the same thing for the week before; if cases seem to be growing, it applies a [delay factor](https://www.microcovid.org/paper/7-basic-method#step-three-delay-factor) to account for the data generally being old by the time it's being used.

However, the idea that there was an update in the last week, much less that the update represents approximately a week of growth over the last number, is a very shaky assumption nowadays.

I wrote an example of this in https://github.com/microCOVID/microCOVID/issues/1685

My tentative belief is that our calculation needs to be more sophisticated, needs to take into account the actual number of days between the different reports made data, and needs to have a more dynamic delay factor based on when the last new data came in, to come up with a better estimate of the number of cases coming in per week.  

I'd want to apply this across the board from all data sources of course.  I do think your `date=` query parameter will come in handy in the process.

This is a pretty high priority on my side to address - with some luck I may have a chance to take a crack at it over the holidays.

In the meantime, I did add code to start detecting these stale conditions and supply a warning in at least the most egregious cases: https://github.com/microCOVID/microCOVID/pull/1686

jeanpaulrsoucy:
Closed due to ending of microCOVID project (https://github.com/microCOVID/microCOVID/discussions/1791).

PR #1819, opened 2023-03-13
temporary update with instructions for using the new interface

octern:
Added specific instructions (for US users at least) for using the new interface, and removed link that makes it sound like the site is being taken down.


src/locales/en.json
```diff
@@ -100,8 +100,8 @@
...
-      "changes_warning_heading": "The microCOVID Project has switched to user-provided prevalence rates",
+      "changes_warning": "<p>microCOVID's data source is no longer providing updates on COVID cases. To use the calculator, you'll need to provide the COVID prevalence in your area. Here's how:</p><p>Go to the red 'Step 1' heading below. Look up <strong>weekly new cases per 100k</strong> in your area and enter it in the first box. Then enter '100,000' in the 'total population' box just below it. The other boxes should be left on their defaults.</p><p><em>In the US, you can get this info at covidactnow.org: Click your state on the map, then your county. The first number, labeled 'weekly new reported cases per 100k', is what you need.</em></p><p>We hope to have an easier interface in place soon. Thank you for bearing with us during this change. Contact info@microcovid.org with any questions.</p>",
```
jaraco, line 103: In a previous edit, @apiology [pointed out](https://github.com/microCOVID/microCOVID/pull/1810#discussion_r1133327823) (more detail in [comms](https://docs.google.com/document/d/1nGZtctzhcZ1CBuWZkSUSavtTyxIwPLeSD9pFVk38ee8/edit#heading=h.bxumnp8dyngm)) that `microCOVID` is a unit of measure (not the site or the calculator) and to use "the microCOVID calculator" to refer to the site/tool.

```suggestion
      "changes_warning": "<p>The microCOVID calculator's data source is no longer providing updates on COVID cases. To use the calculator, you'll need to provide the COVID prevalence in your area. Here's how:</p><p>Go to the red 'Step 1' heading below. Look up <strong>weekly new cases per 100k</strong> in your area and enter it in the first box. Then enter '100,000' in the 'total population' box just below it. The other boxes should be left on their defaults.</p><p><em>In the US, you can get this info at covidactnow.org: Click your state on the map, then your county. The first number, labeled 'weekly new reported cases per 100k', is what you need.</em></p><p>We hope to have an easier interface in place soon. Thank you for bearing with us during this change. Contact info@microcovid.org with any questions.</p>",
```

jaraco, line 103: An observation: I don't think it's necessary to prescribe "per 100,000" and then to enter "100,000". It's also possible to enter total cases and total population. I do think it's useful to point out that it's possible to do it either way (and between the form and the guidance, it hints that either is possible).

jaraco (approved):

PR #1821, opened 2023-03-13
Add instructions to the prevalence inputs.

jaraco:

PR #1822, opened 2023-03-14
Assume population of 100,000 in manual input.

apiology:


src/locales/en.json
```diff
@@ -100,7 +100,7 @@
...
+      "changes_warning": "<p>The microCOVID calculator's data source is no longer providing updates on COVID cases. To use the calculator, you'll need to provide the COVID prevalence in your area. Here's how:</p><p>Go to the red 'Step 1' heading below. Look up <strong>weekly new cases per 100k</strong> in your area and enter it in the first box. The other boxes should be left on their defaults.</p><p><em>In the US, you can get this info at <0>covidactnow.org</0>: Click your state on the map, then your county. The first number, labeled 'weekly new reported cases per 100k', is what you need.</em></p><p>We hope to have an easier interface in place soon. Thank you for bearing with us during this change. Contact info@microcovid.org with any questions.</p>",
```
apiology, line 103: @jaraco: You'll need to recreate this change in #1821 if I beat you to the merge.

octern (approved):

src/locales/en.json
jaraco: Looks like it merged fine.

jaraco:
I'm not a fan of this change. Why not leave the input but have it default to 100000? What if the stats are only reported for total population? Then, the user has to multiply by 100000/total_pop.

PR #1825, opened 2023-03-16
Use percentFullyVaccinated for UI warning

### LINKED ISSUE #1817: Warning appears for vacination even if 'percent of population vaccinated' is supplied
image

_Originally posted by @jaraco in https://github.com/microCOVID/microCOVID/issues/1810#issuecomment-1464978424_

jaraco:
Fixes #1817.


jaraco:
I'm not sure if this change is the right one. It does cause the warning to go away when vaccination percentage is applied, but the value of the vaccination doesn't seem to have any effect on the calculator, so I wonder if the vaccine percentage should be eliminated altogether.

jaraco:
@justinhaaheim Can you take a peek and see if this change makes sense from a technical point of view (is this variable not the correct one to use for the warning message)? Also, according to the failed check, this line isn't covered by any tests. Could you devise a test that covers this behavior or do you recommend to bypass the failed check?

jaraco:
This change will likely be superseded by an effort to remove percent vaccinated from the inputs (#1826).

Issue #1826, opened 2023-03-18
Manual prevalence entry v1 proposal


octern:
### **Goal of this proposal**
- Make the site sustainably usable for users who are willing to look up and manually input **one** piece of data about local COVID rates. 
Note: this is based on several discussions, primarily on #disc-code and [these meetings](https://docs.google.com/document/d/1Wf4CVQh6ImoaapoqYeQU_3Yw2CJFbPVUpDqn5AIhEd8/edit#)

### **Proposed changes:**

* Blue infobox
   - [x] Move the specific data entry instructions from the blue infobox into a box immediately under the "step 1" header (#1821, #1827)
* Reported cases field:
	- [ ] Label as "weekly new cases / 100k population". 
	- [x] remember past values, the same way we do now
	- [x] No default value; if there's not a past value leave it blank so the user has to enter something
* Population field:
	- [x] If the box would be blank, set it to 100,000 (#1830)
	- [x] Remember past values, the same way we do now (If the past value isn't 100k, the user will probably also have a saved value for total cases, not per 100k)
	- [ ] If the value is not 100,000 show text with a warning
* Change-over-past-week field
	- [x] We do not expect users to be able to obtain this value with a reasonable amount of effort. 
	- [ ] #1834
	- [ ] **for the present version** at least, hide the field (visually only, so that the adjusted prevalence formula still has access to the parameter it expects)
* Test positivity rate field
	- [x] Users could look this up but it may be an unreasonable amount of work compared to the improvement in accuracy. 
	- [x] **for the present version** wipe any saved values in this field and set to 10% (a cautious, ie high, estimate) (#1830)
	- [ ] **for the present version**, at least, hide the field (visually only, so the ) adjusted prevalence formula still has access to the parameter it expects)
* Vaccine rate field
	- [x] This currently has trivial effects on adjusted prevalence
	- [x] Carry forward past values, or by default set to zero -- it doesn't matter as long as there's some value in it
	- [ ] Hide the field (visually only, so that the adjusted prevalence formula still has access to the parameter it expects)

### Technical issues to deal with:
* Under some circumstances past values can override the values we try to set as defaults (see #1823 )
* ...?

### Process
The current agreement was to develop this on its own branch ([manual-entry-v1](https://github.com/microCOVID/microCOVID/tree/manual-entry-v1)), relying on the Netlify builds for testing (pending resolution of #1828). Once we have all the changes in place, there may or may not be interest in releasing behind a feature flag. 

[...]

PR #1827, opened 2023-03-19
Remove instructions from the blue box

jaraco:
Now that instructions are inline with the form (#1821), remove the instructions from the blue box (ref #1826).


apiology (commented):
Looks right if @octern is good with it

octern:
Could you edit the blue box so that the first paragraph ends with "... prevalence in your area. The "Step 1" heading below has simple instructions for looking this up!"

jaraco:
> Could you edit the blue box so that the first paragraph ends with "... prevalence in your area. The "Step 1" heading below has simple instructions for looking this up!"

Done in 79ae555.

octern (approved):

PR #1829, opened 2023-03-19
Assume population of 100,000 in manual input.

jaraco:
This reverts commit b5550401b42baf79791d96f3550f51a200f0e16d, restoring f14ad0b9fe9c78b46f81405b9482bdf5a754d9ff.

This change is a re-introduction of #1822.


jaraco:
I realized after proposing this change that the approach needs to be revisited because the latest proposal is to continue to keep the field but default to 100k and warn if it's not.

Issue #1828, opened 2023-03-19
Tests and previews unavailable for branch-based workflow


jaraco:
In #1826, the active maintainers have proposed a process to push changes to a branch. For that, I created the [manual-entry-v1 branch](https://github.com/microCOVID/microCOVID/tree/manual-entry-v1). In #1827, I attempted to use that process to submit a change, but quickly realized that only main is configured to run tests and serve netlify previews, making the pull request not particularly useful for validating the change.

![image](https://user-images.githubusercontent.com/308610/226182012-c032763f-0cfd-49b2-8898-a6fbaddaa6b7.png)

To make this workflow viable, someone will need to configure that branch to have similar behaviors as the main branch.

Alternatively, we could consider reverting to the simpler but less safe direct-to-main workflow.

jaraco:
Fixed in #1833.

PR #1830, opened 2023-03-19
Default the total population to 100k and positivity to 10%

jaraco:
Ref #1826


jaraco:
Due to #1828, retargeting to `main`.

apiology (commented):
I don't really know enough to comment on the code being sufficient without testing personally, but this certainly seems a simple way to do it (too simple? idk). 

I'd defer to @octern's testing for the final OK.

octern:
LGTM. Looks like this only has an effect if the user doesn't have a stored value, which is fine.

octern (approved):
LGTM, should fail gracefully if the user has any values saved.

PR #1832, opened 2023-03-22
DO NOT MERGE - demo of adding a new feature to a dev branch (manual-entry-v1)

justinhaaheim:


apiology:
@jaraco: Netlify works on this now with the updated settings I noted in #1828

jaraco:
Thanks for this example. With #1833 and apiology's changes to netlify config, this approach could be viable. We'll want to decide as a group what workflow we want.

PR #1833, opened 2023-03-24
Run tests on all branches.

jaraco:
Ref #1828


jaraco:
I'm not at all confident this change will have the desired effect. I do suspect it will cause the CI checks to run for the indicated branch, but I suspect it won't affect the netlify previews, which are the most important thing for a reviewer to review.

apiology (approved):
Should this be on the 'push' key as well?

apiology:
> I'm not at all confident this change will have the desired effect. I do suspect it will cause the CI checks to run for the indicated branch, but I suspect it won't affect the netlify previews, which are the most important thing for a reviewer to review.

Ah, yeah, you're right.

I see docs here: https://docs.netlify.com/site-deploys/overview/#branches-and-deploys

Current config:

image

New config after I tweaked:

image

jaraco:
> Should this be on the 'push' key as well?

I don't really have an opinion. I don't know if this value affects forks or just the main fork. I've not configured this setting on my other repos, so I don't have much experience with it. I'm unsure what the motivations were for specifying the branches.

Looking at the history, there's been a lot of toil around these values, adding and removing branch names every time a new branch becomes active or inactive.

This makes me think we should just keep it simple and remove the constraint and test all branches. I'll update the PR accordingly.

apiology (approved):

PR #1837, opened 2023-04-24
#1826 - Manual prevalence entry v1 proposal

CloudByteKS:
- Please see [here](https://github.com/microCOVID/microCOVID/issues/1826) for source of changes.
- Hide specific fields from view (but keep in the page)
- Hardcode some of the newly hidden values so that all previous values are "wiped out" and are not affecting the calculation
- Show a Warning if Population field is not exactly 100k.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,9 @@
 # See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
+# environment
+.env
+.idea
+
 # dependencies
 /node_modules
 /.pnp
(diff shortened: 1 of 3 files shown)


octern:
@schultz96 Thank you! A few requested changes:
1. The new cases box currently doesn't allow decimal points. Please have it allow them.
3. The total population box ignores commas but the warning message for it doesn't. Please change so that there's not a warning if someone enters `100,000`
4. Please remove the sentence "The other boxes can be left on their defaults."

CloudByteKS:
> @schultz96 Thank you! A few requested changes:
> 
> 1. The new cases box currently doesn't allow decimal points. Please have it allow them.
> 2. The total population box ignores commas but the warning message for it doesn't. Please change so that there's not a warning if someone enters `100,000`
> 3. Please remove the sentence "The other boxes can be left on their defaults."

I've updated the code based on your feedback. Let me know if you encounter any other issues or problems!

octern (approved):

PR #1841, opened 2023-05-06
Implement underreporting bump of 4x.

### LINKED ISSUE #1840: Update prevalence multiplier from 2x to 8x
Per a conversation and [follow-up in Slack](https://microcovid.slack.com/archives/C019NKP5L9G/p1681310958324479), to account for under-reporting of cases, we propose to add another 4x multiplier to the model when calculating prevalence.

jaraco:
- Rename 'prevalance_ratio' to 'underreporting_factor' for clarity.
- Increase the underreporting factor by 4x. Ref #1840.
- Implement underreporting factor of 4x. Fixes #1840.


jaraco:
I don't plan to write a test to cover this change. We might consider asking Kenny to add one.

